PR #3, opened 2020-07-23
Adds metrics for pinot request handler query execution time

kotharironak:


query-service-impl/build.gradle.kts
```diff
@@ -35,6 +35,7 @@ dependencies {
   }
   implementation("org.slf4j:slf4j-api:1.7.30")
   implementation("com.typesafe:config:1.3.2")
+  implementation("org.hypertrace.core.serviceframework:platform-metrics:0.1.2")
```
tim-mwangi, line 38: 0.1.3

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -109,8 +122,11 @@ public void handleRequest(
       LOG.debug("Trying to execute PQL: [ {} ] by RequestHandler: [ {} ]", pql, this.getName());
     }
     final PinotClient pinotClient = pinotClientFactory.getPinotClient(this.getName());
+    Context timerContext = pinotQueryExecutionTimer.time();
     try {
       final ResultSetGroup resultSetGroup = pinotClient.executeQuery(pql.getKey(), pql.getValue());
+      timerContext.stop();
```
tim-mwangi, line 128: finally?

query-service-impl/build.gradle.kts
kotharironak: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
kotharironak: I have added into exception block below. As part of this metric, would like to measure `executeQuery`.
```
catch (Exception ex) {
      // stop any timer context
      timerContext.stop();
      // Catch this exception to log the Pinot SQL query that caused the issue
      LOG.error("An error occurred while executing: {}", pql.getKey(), ex);
      // Rethrow for the caller to return an error.
      throw ex;
    }
```

kotharironak: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotUtils.java
```diff
@@ -7,4 +7,8 @@ public static String getZkPath(String zkBasePath, String pinotClusterName) {
         ? zkBasePath + pinotClusterName
         : zkBasePath + "/" + pinotClusterName;
   }
+
+  public static String getMetricName(String name) {
+    return name.replaceAll("[^a-zA-Z0-9_]", ".");
```
inespo, line 12: wonder if this should go to the platform-metrics and only allow the alphaNumeric value.

kotharironak: Right now, this one is a special use case, other places we are using constant names. yes, but we can consider that. will chat with on this.

PR #4, opened 2020-07-28
adds supports for filtering on bytes ID columns

kotharironak:
There are few strings IDs fields are maps to bytes column in pinot, as part of this PR,
- adds support for handling filtering query on such fields by type conversion (e.g parent_span_id != 'null')
- fails fast on invalid string input for such fields if it fails to convert to valid hex string
- re-factored init steps for pinot query conversion tests

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -30,396 +30,395 @@ import org.slf4j.LoggerFactory;
  */
 class QueryRequestToPinotSQLConverter {
 
-  private static final Logger LOG = LoggerFactory.getLogger(QueryRequestToPinotSQLConverter.class);
+    private static final Logger LOG = LoggerFactory.getLogger(QueryRequestToPinotSQLConverter.class);
 
-  private static final String QUESTION_MARK = "?";
-  private static final String REGEX_OPERATOR = "REGEXP_LIKE";
-  private static final String MAP_VALUE = "mapValue";
-  private static final int MAP_KEY_INDEX = 0;
-  private static final int MAP_VALUE_INDEX = 1;
+    private static final String QUESTION_MARK = "?";
+    private static final String REGEX_OPERATOR = "REGEXP_LIKE";
+    private static final String MAP_VALUE = "mapValue";
+    private static final int MAP_KEY_INDEX = 0;
+    private static final int MAP_VALUE_INDEX = 1;
 
-  private ViewDefinition viewDefinition;
-  private Joiner joiner = Joiner.on(", ").skipNulls();
+    private ViewDefinition viewDefinition;
+    private Joiner joiner = Joiner.on(", ").skipNulls();
 
-  QueryRequestToPinotSQLConverter(ViewDefinition viewDefinition) {
-    this.viewDefinition = viewDefinition;
-  }
-
-  Entry<String, Params> toSQL(
-      QueryContext queryContext, QueryRequest request, LinkedHashSet<Expression> allSelections) {
-    Params.Builder paramsBuilder = Params.newBuilder();
-    StringBuilder pqlBuilder = new StringBuilder("Select ");
-    String delim = "";
-
-    // Set the DISTINCT keyword if the request has set distinctSelections.
-    if (request.getDistinctSelections()) {
-      pqlBuilder.append("DISTINCT ");
+    QueryRequestToPinotSQLConverter(ViewDefinition viewDefinition) {
+        this.viewDefinition = viewDefinition;
     }
 
-    // allSelections contain all the various expressions in QueryRequest that we want selections on.
-    // Group bys, selections and aggregations in that order. See RequestAnalyzer#analyze() to see
-    // how it is created.
-    for (Expression expr : allSelections) {
-      pqlBuilder.append(delim);
-      pqlBuilder.append(convertExpression2String(expr, paramsBuilder));
-      delim = ", ";
-    }
-
-    pqlBuilder.append(" FROM ").append(viewDefinition.getViewName());
-
-    // Add the tenantId filter
-    pqlBuilder.append(" WHERE ").append(viewDefinition.getTenantIdColumn()).append(" = ?");
-    paramsBuilder.addStringParam(queryContext.getTenantId());
-
-    if (request.hasFilter()) {
-      pqlBuilder.append(" AND ");
-      String filterClause = convertFilter2String(request.getFilter(), paramsBuilder);
-      pqlBuilder.append(filterClause);
-    }
-
-    if (request.getGroupByCount() > 0) {
-      pqlBuilder.append(" GROUP BY ");
-      delim = "";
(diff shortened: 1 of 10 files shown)


kotharironak:
Verified locally running query service. Will verify with locally running HT.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/StringToValueConverter.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.protobuf.ByteString;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang.StringUtils;
+import org.hypertrace.core.query.service.api.Value;
+import org.hypertrace.core.query.service.api.ValueType;
+
+/**
+ * Converter to convert any string value to requested {@link Value} of {@link ValueType}
+ * */
+public class StringToValueConverter implements ToValueConverter<String> {
```
buchi-busireddy, line 12: Rename to `StringToBytesValueConverter`?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -164,16 +171,65 @@ private String convertFilter2String(Filter filter, Params.Builder paramsBuilder)
...
+   */
+  private Expression handleValueConversionForLiteralExpression(Expression lhs, Expression rhs) {
+    if (!(lhs.getValueCase().equals(COLUMNIDENTIFIER) && rhs.getValueCase().equals(LITERAL))) {
+      return rhs;
+    }
+
+    Expression newRhs = rhs;
+    String lhsColumnName = lhs.getColumnIdentifier().getColumnName();
+
+    if (viewDefinition.isBytesColumn(lhsColumnName)) {
```
buchi-busireddy, line 200: I think we should rather pass the column type to the `ToValueConverter` and let the implementation take care what to do for string vs bytes columns. We can avoid this if check and special logic for Bytes with that.

the converter instance will take care of bytes and other cases so why do we need to check for `isBytesColumn` explicitly?

kotharironak: Sounds good. I will make function generic one.

kotharironak: Sounds good, will make it generic. I was avoiding StringToString conversion as default types is a column and re-creation of the Value object.

Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -348,6 +405,8 @@ private String convertLiteralToString(LiteralConstant literal, Params.Builder pa
         paramsBuilder.addDoubleParam(value.getDouble());
         break;
       case BYTES:
+        ret = QUESTION_MARK;
+        paramsBuilder.addBytesStringParam(Hex.encodeHexString(value.getBytes().toByteArray()));
```
tim-mwangi, line 409: Careful with `Hex.encodeHexString()` especially if you need to access the ByteBuffer later on.(Which you do not need to over here)

tim-mwangi: Also what happens when this is `'null'`

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/StringToValueConverter.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private StringToValueConverter() {
+  }
+
+  public Value convert(String value, ValueType valueType) throws Exception {
+    Value.Builder valueBuilder = Value.newBuilder();
+    switch (valueType) {
+      case BYTES:
+        String outValue = (StringUtils.isEmpty(value) ||
+            value.equals("null") ||
```
tim-mwangi, line 25: Is this correct? `null` is not equal to `''`

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewDefinition.java
```diff
@@ -35,6 +37,10 @@ public static ViewDefinition parse(Map<String, Object> config, String tenantColu
       mapFields.addAll(mapFieldsList);
     }
 
+    // get bytes fields
+    final Set<String> bytesFields = new HashSet<>(
+        (List<String>) config.getOrDefault("bytesFields", new ArrayList<>()));
```
tim-mwangi, line 42: List.of() instead of new arraylist

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
buchi-busireddy: Actually, would this ever even happen? None of our attributes are typed as `BYTES` right?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/StringToValueConverter.java
buchi-busireddy: Should be about right in this case since there shouldn't be any span ids or trace ids which are `bytes[0]` right? any bytes field for that matter.

kotharironak: Yes. this is fine as we are converting default string values to a default value of BYTES in pinot which `byte array of length 0`.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
kotharironak: yes. we don't have BYTES exposed at the gateway layer. The default value for bytes in proto is `bytes[0]`. So, this will be fine if we ever expose it.

kotharironak:
@buchi-busireddy I will have to update the application.conf file as Samarth has updated it. will scan for other ID attributes that need to update in other handlers too before merging this one.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/StringToValueConverter.java
tim-mwangi: ok.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewDefinition.java
kotharironak: Done.

kotharironak:
@buchi-busireddy I have updated newly added handlers for BYTES fields.

PR #6, opened 2020-07-30
removing service configs from helm values.yaml (local and ht helm)

samarth-gupta-traceable:
sample PR for removing service configs from helm values.yaml

have taken service handlers from `hypertrace-helm` also and added in `application.conf`

### DIFF
--- a/helm/templates/query-service-config.yaml
+++ b/helm/templates/query-service-config.yaml
@@ -14,7 +14,7 @@ data:
           connectionString = "{{ .Values.queryServiceConfig.data.zookeeperConnectionString }}"
         }
       ]
-      {{- if .Values.handlers }}
+
       queryRequestHandlersConfig = [
       {{- range .Values.handlers }}
         {
@@ -27,5 +27,5 @@ data:
         }
       {{- end }}
       ]
-      {{- end }}
+
     }
--- a/helm/templates/query-service-config.yaml
+++ b/helm/templates/query-service-config.yaml
@@ -14,7 +14,7 @@ data:
           connectionString = "{{ .Values.queryServiceConfig.data.zookeeperConnectionString }}"
         }
       ]
-
+      {{- if or .Values.handlers .Values.extraHandlers }}
       queryRequestHandlersConfig = [
       {{- range .Values.handlers }}
         {
@@ -27,5 +27,5 @@ data:
         }
       {{- end }}
       ]
-
+      {{- end }}
     }
--- a/query-service/src/main/resources/configs/common/application.conf
+++ b/query-service/src/main/resources/configs/common/application.conf
@@ -20,8 +20,11 @@ service.config = {
           viewName = rawTraceView
           fieldMap = {
             "TRACE.id": "trace_id",
+            "TRACE.name": "transaction_name",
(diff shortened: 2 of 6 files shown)


query-service/src/main/resources/configs/common/application.conf
```diff
@@ -38,18 +41,201 @@ service.config = {
...
+            "SERVICE.errorCount": "error_count",
+            "SERVICE.exceptionCount": "exception_count",
+            "SERVICE.numCalls": "num_calls",
+            "API.isExternal" : "api_is_external"
+          }
+        }
+      }
+    }
+    {
+      name = api-traces-view-handler
```
kotharironak, line 94: Is all handler available in hypertrace-core?

samarth-gupta-traceable: @kotharironak  for now we are proceeding with config changes as required for `hypertrace`. 
 pls see slack channel #ht-deploy-launch.

buchi-busireddy: No @kotharironak

helm/templates/query-service-config.yaml
```diff
@@ -14,7 +14,7 @@ data:
           connectionString = "{{ .Values.queryServiceConfig.data.zookeeperConnectionString }}"
         }
       ]
-      {{- if .Values.handlers }}
```
ravisingal, line 17: why this check is removed? now this config will have `queryRequestHandlersConfig = []` which will override the queryRequestHandlersConfig in default application.conf file.

samarth-gupta-traceable: @ravisingal  i had to remove this since when I tested with saas `helm-values` locally , complete `queryRequestHandlersConfig[]` itself was missing from generated configmap `query-service-config` . I think that is coz `{{- if .Values.handlers }}` returns false for empty array `handlers[]` .  So i removed this if block.
But you are correct this will also override default queryRequestHandlersConfig with empty once. 

Do you have some way using which both the above issues can be addressed ? 
Also I think we might have to provide the default handlers that I removed from this repo `values.yaml` in saas `helm-values` also ...

samarth-gupta-traceable: @buchi-busireddy  I tested locally and I dont think 2 `service.config` from different application.conf are getting merged at `queryRequestHandlersConfig`  level.

For example if `queryRequestHandlersConfig` in `/app/resources/configs/common/application.conf` have data  but `queryRequestHandlersConfig` in `/app/resources/configs/query-service/application.conf` is empty , then query-service is starting with empty configuration and not merging both of them.

ravisingal: @samarth-gupta-traceable Add the condition as follow:
```
{{- if or .Values.handlers .Values.extraHandlers }}
```

samarth-gupta-traceable: @ravisingal  fixed the above as per your suggestion. 

Also created PR to add `handlers` removed from this repo helm chart values.yaml to `helm-values` values.yaml
https://github.com/Traceableai/helm-values/pull/130

we will need to merge above PR first before merging this i guess.
cc @buchi-busireddy

PR #8, opened 2020-08-10
provides a way to override common properties as env

kotharironak:


query-service/src/main/resources/configs/common/application.conf
```diff
@@ -3,10 +3,12 @@ service.name = query-service
 service.port = 8090
 service.admin.port = 8091
 service.config = {
+  tenantColumnName = ${?TENANT_COLUMN_NAME}
```
ravisingal, line 6: what is default value?

kotharironak: There is no default for this, we want it to explicitly set.

buchi-busireddy: @ravisingal I think we can set the default value directly to `tenant_id` here instead of ENV variable.

kotharironak: sounds good here as we have already hardcoded this in schema, we can do that.

kotharironak: Done.

PR #10, opened 2020-08-13
Changing the slow queries format to be easily usable for debugging.

buchi-busireddy:
You can either take the query to reconstruct the proto or get the full sql
so that you can debug it much more quickly.
Also did a bunch of code cleanup.

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
@@ -3,6 +3,8 @@ package org.hypertrace.core.query.service.pinot;
 import com.codahale.metrics.Timer;
 import com.codahale.metrics.Timer.Context;
 import com.google.common.base.Preconditions;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -113,7 +118,7 @@ public void handleRequest(
       QueryContext queryContext,
       QueryRequest request,
       QueryResultCollector<Row> collector,
-      RequestAnalyzer requestAnalyzer) {
+      RequestAnalyzer requestAnalyzer) throws InvalidProtocolBufferException {
```
tim-mwangi, line 121: This would rarely happen. Very low chance actually since this is in memory proto request. We can do try catch inside

buchi-busireddy: Fair point. Fixed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -125,16 +130,20 @@ public void handleRequest(
...
       }
       // need to merge data especially for Pinot. That's why we need to track the map columns
       convert(resultSetGroup, collector, requestAnalyzer.getSelectedColumns());
-      long requestTimeMs = System.currentTimeMillis() - start;
-      if (requestTimeMs > SLOW_REQUEST_THRESHOLD_MS) {
-        LOG.warn("Query Execution time: {} millis\nQuery Request: {}", requestTimeMs, request);
+
+      try {
+        long requestTimeMs = System.currentTimeMillis() - start;
+        if (requestTimeMs > SLOW_REQUEST_THRESHOLD_MS) {
```
inespo, line 142: Can we make this threshold configurable? for debugging purpose, in some use cases 3s may not be acceptable threshold. it's okay to put in different PR though, since it is outside the scope of this review.

buchi-busireddy: Yeah, thought about it. The current way code is structured, it's not straightforward so deferred it but I'm planning to address in future PRs when I get time.

PR #11, opened 2020-08-14
Use T-digest percentile since regular percentile in Pinot is super expensive.

buchi-busireddy:
* This PR also makes the percentile function configurable for Pinot request handlers. 
* Using new metrics library and report Pinot query latency with handler tag.
* This also pushes the tenantColumnName config to the request handler level since that's where it should ideally belong.
* Another change handled is making the slowQueryThreshold configurable, per handler.


query-service-client/build.gradle.kts
```diff
@@ -12,5 +12,5 @@ dependencies {
   // Logging
   implementation("org.slf4j:slf4j-api:1.7.30")
   // Config
-  implementation("com.typesafe:config:1.3.2")
```
tim-mwangi, line 15: while you are can you upgrade `io.grpc` dependencies to `1.31.1` and grpc utils to `0.1.4` and then get rid of the snyk ignore message? Keep the .snyk file but remove the vulnerability info. See https://github.com/hypertrace/java-grpc-utils/pull/4

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -18,6 +17,7 @@ service.config = {
       type = pinot
       clientConfig = zookeeper
       requestHandlerInfo = {
+        tenantColumnName = tenant_id
```
tim-mwangi, line 20: Why did you move this into the request handler info? Seems wrong to me since it opens up more config bug opportunities.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -118,7 +149,7 @@ public void handleRequest(
       QueryContext queryContext,
       QueryRequest request,
       QueryResultCollector<Row> collector,
-      RequestAnalyzer requestAnalyzer) {
+      RequestAnalyzer requestAnalyzer) throws Exception {
```
tim-mwangi, line 152: Who is throwing the exception and can the be wrapped into a runtime exception?

query-service-client/build.gradle.kts
buchi-busireddy: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Removed it now and cleaned up the code a bit.

query-service/src/main/resources/configs/common/application.conf
buchi-busireddy: Because this is the right design. 
Imagine there are different data stores are present and there are handlers for each store, table combo. Do you want to dictate the same column name across all tables and stores? The tenant_id column belongs at the handler level so that they can choose whatever they want to call it as.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceImplConfig.java
```diff
@@ -34,16 +28,12 @@ public void setQueryRequestHandlersConfig(List<Config> queryRequestHandlersConfi
...
-    this.tenantColumnName = tenantColumnName;
-  }
-
   public static class RequestHandlerConfig {
 
     private String name;
     private String type;
     private String clientConfig;
-    private Map<String, Object> requestHandlerInfo;
+    private Config requestHandlerInfo;
```
aaron-steinfeld, line 36: Not that the old interface was any better, but would it make sense to make this a real interface rather than passing around an untyped map of config? Not sure where the assumptions about config keys are made, but moving that to as close to the file read as possible makes it less error prone and more debuggable.

query-service-impl/src/test/resources/application.conf
```diff
@@ -42,6 +44,9 @@ service.config = {
       type = pinot
       clientConfig = zookeeper
       requestHandlerInfo = {
+        tenantColumnName = tenant_id
+        slowQueryThresholdMs = 100
```
tim-mwangi, line 48: Why is this being configured to a request handler? Isn't a slow request a slow request regardless of where it's being executed?

query-service/src/main/resources/configs/common/application.conf
tim-mwangi: I'd understand there were multiple data sources but we are not there yet. Also if you wanna go this route you will need to make changes to helm-values repo.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceImpl.java
```diff
@@ -30,41 +28,25 @@ public QueryServiceImpl(QueryServiceImplConfig config) {
...
-                  config
-                      .getRequestHandlerInfo()
-                      .get(PinotBasedRequestHandler.VIEW_DEFINITION_CONFIG_KEY),
-              tenantColumnName));
-      RequestHandlerRegistry.get()
-          .register(
-              config.getName(),
-              new RequestHandlerInfo(
-                  config.getName(), PinotBasedRequestHandler.class, requestHandlerInfoConf));
+      boolean registered = RequestHandlerRegistry.get().register(config.getName(),
```
tim-mwangi, line 44: This is hard to test?

query-service-impl/src/test/resources/application.conf
buchi-busireddy: Flexibility. Applying the same slow query threshold for all tables and data stores isn't ideal and would give lot of false positives. Since each view/table could be different and query patterns on them could be different, this would allow that flexibility.
There is a default anyways that's applicable for all and it's not mandatory to override. In fact, i'm not setting this at handler level for actual service. This is a test.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceImpl.java
buchi-busireddy: Currently the way PinotClient is created and first connections are made, yes, it's not straight forward. Maybe in a next PR.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceImplConfig.java
buchi-busireddy: The keys of this config are assumed in the implementations of `RequestHandler`.
The reason this is a generic untyped config is, each handler could look differently and might have different keys for configs. For example: A pinot RequestHandler vs Druid RequestHandler config could look totally different and might need different set of keys.
Having said that, there could be ways to improve this further. I'll think about this and we can fix in future PRs. Given that I haven't changed the conf files now, this shouldn't be a big regression.

aaron-steinfeld: Agreed it's not a regression and that it needs to be supported per-handler type. The same place we translate to the appropriate handler, we could translate to the appropriate config. That could fairly easily happen inside the handler implementation too, I just prefer to separate concerns and the quicker a `Config` gets converted to a typed object, the easier it is to trace back issues. Nothing need be done right now, for sure.

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryServiceImplTest.java
```diff
@@ -65,7 +65,6 @@ public void testBlankTenantColumnNameThrowsException() {
 
     // whitespace tenant id column name
     QueryServiceImplConfig queryServiceConfig2 = new QueryServiceImplConfig();
-    queryServiceConfig2.setTenantColumnName("   ");
```
tim-mwangi, line 68: These tests are not valid anymore, are they? Also where are the tests to check for the empty tenant id now.

query-service/src/main/resources/configs/common/application.conf
buchi-busireddy: Replied to you separately for this.

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryServiceImplTest.java
buchi-busireddy: you're right. Removed. Thanks for pointing this.

buchi-busireddy: `org.hypertrace.core.query.service.pinot.PinotBasedRequestHandlerTest#testNullTenantIdQueryRequestContextThrowsNPE`

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -82,17 +93,40 @@ public String getName() {
...
+      this.percentileAggFunction = config.getString(PERCENTILE_AGGREGATION_FUNCTION_CONFIG);
+    }
+    LOG.info("Using {} function for percentile aggregations of handler: {}",
+        this.percentileAggFunction, name);
+
+    this.request2PinotSqlConverter =
+        new QueryRequestToPinotSQLConverter(viewDefinition, this.percentileAggFunction);
+
+    if (config.hasPath(SLOW_QUERY_THRESHOLD_MS_CONFIG)) {
+      this.slowQueryThreshold = config.getInt(SLOW_QUERY_THRESHOLD_MS_CONFIG);
```
inespo, line 118: thanks! very nice

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -282,6 +285,8 @@ private String convertExpression2String(Expression expression, Params.Builder pa
         // COUNT(*) for ORDER BY
         if (functionName.equalsIgnoreCase("COUNT")) {
           return functionName + "(*)";
+        } else if (functionName.startsWith(PERCENTILE_PREFIX) && !PERCENTILE_PREFIX.equals(functionName)) {
```
inespo, line 288: I think abstracting the percentile at this level is correct. This allows us to potentially change storage without changing the QueryService API.
Should we allow both Percentile and ApproxPercentile though as a function? 
We can add it later as long as we can configure function -> implementation

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandlerTest.java
```diff
@@ -44,6 +50,80 @@ public void setUp() {
...
+                .build(),
+            Set.of(), Set.of("Trace.start_time_millis", "Trace.end_time_millis", "Trace.duration_millis",
+                "Trace.id", "Trace.tags"));
+        Assertions.assertTrue(cost.getCost() >= 0.0d && cost.getCost() < 1.0d);
+      }
+    }
+  }
+
+  @Test
+  public void testCanHandleNegativeCase() {
```
inespo, line 102: Nice!

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
buchi-busireddy: The value of config lets you add either Percentile or Approx. That's the config indeed :)

PR #15, opened 2020-08-31
Handling the empty bracket scenario for bytes filter

kotharironak:
This PR handles the filter scenario for bytes type,

```
childFilter {
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "EVENT.id"
        }
      }
      operator: EQ
      rhs {
        literal {
          value {
            string: "{}"
          }
        }
      }
    }
  }
      }
```

It adds '{}' as a valid list of null values for bytes field


kotharironak:
Verified locally using hypertrace:
Screenshot 2020-08-31 at 5 27 09 PM

PR #17, opened 2020-09-09
Adding a feature to add column filters to Pinot query handlers.

buchi-busireddy:
This feature will be useful when there are two different Pinot views which serve the same set of attributes but they have different sets of data.
As an example, you could have all spans in one table but you can also have just entry spans in a different Pinot table, which is optimized for entry span queries. Without this feature, there is no way to select the correct view for the span queries in this example but with this feature, you can select the optimized view for entry span queries.


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
```diff
@@ -56,6 +56,58 @@ private Params(
...
+    }
+    if (!doubleParams.equals(params.doubleParams)) {
+      return false;
+    }
+    return bytesStringParams.equals(params.bytesStringParams);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = integerParams.hashCode();
```
tim-mwangi, line 102: Why are you not using Objects methods?

query-service-impl/src/test/resources/log4j2.properties
```diff
@@ -6,3 +6,6 @@ appender.console.layout.type=PatternLayout
 appender.console.layout.pattern=%d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %-5level %c{1.} - %msg%n
 rootLogger.level=INFO
 rootLogger.appenderRef.stdout.ref=STDOUT
+
+logger.hypertrace.level = DEBUG
```
tim-mwangi, line 10: rollback to info

tim-mwangi: oh never mind. this is impl.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
```diff
@@ -56,6 +56,58 @@ private Params(
     return bytesStringParams;
   }
 
+  @Override
+  public String toString() {
```
tim-mwangi, line 60: Should be easy to unit test these, no?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,6 +185,130 @@ public QueryCost canHandle(
     return queryCost;
   }
 
+  /**
+   * Method to check if the given ViewColumnFilter matches the given query filter. A match here
+   * means the view column is superset of what the query filter is looking for, not an exact match.
+   */
+  private boolean doesFiltersMatch(ViewColumnFilter viewColumnFilter, Filter queryFilter) {
```
tim-mwangi, line 192: doFiltersMatch?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -189,6 +358,46 @@ public void handleRequest(
...
+      List<Filter> newFilters = filter.getChildFilterList().stream()
+          .map(f -> removeViewColumnFilter(f, columnName))
+          .filter(Objects::nonNull).collect(Collectors.toList());
+      if (newFilters.size() == 1) {
+        return newFilters.get(0);
+      } else {
+        return Filter.newBuilder(filter).clearChildFilter().addAllChildFilter(newFilters).build();
+      }
+    } else {
+      if (filter.getLhs().getColumnIdentifier().getColumnName().equals(columnName)) {
```
tim-mwangi, line 394: What if columnIdentifier is not set? Can that even happen with proto?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,6 +185,130 @@ public QueryCost canHandle(
...
+    }
+    return expressionValues;
+  }
+
+  /**
+   * Returns a map from column name to the Filter that's received for that column in the
+   * given query.
+   */
+  @Nonnull
+  private Map<String, Filter> getQueryFilterMap(QueryRequest request) {
```
tim-mwangi, line 289: You are not caring if it's an OR or AND filter?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
buchi-busireddy: Any advantages that gives specifically? I can check that.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
buchi-busireddy: It's only used for logging, hence didn't bother to add the unit test. Equals is unit tested though.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Will do.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: ideally not but I can add a requirement here so that we don't run into NPE.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: I've thought about it. If we handle `OR` and `AND`, the behavior wouldn't change for `AND` but we can actually optimize for `OR` scenario. However, I wanted to avoid over-engineering and keep the logic simple for now but I could be convinced to optimize that though :)

aaron-steinfeld: Haven't dove into this enough to know if I'm misunderstanding the comment, but OR vs AND seems like it should matter.

A view filtered to `FOO = x` should match a query with `FOO = x AND BAR = y` but can't match `FOO = X OR BAR = y` (since there could be spans with `BAR = Y` not present in this view). Is that distinction accounted for somewhere?

buchi-busireddy: @aaron-steinfeld Good point, and that case is not handled with the current change. I was originally debating whether that validation is responsibility of the above layers or the view handler. Handling it here would let the view filter attribute to be treated like any other attribute so I like what you're suggesting.

It sounds like we have to check the entire query tree to make sure there is no `OR` in the hierarchy from the matching view filter node until the root. Is that what you're thinking too?

buchi-busireddy: actually, nm about my second sentence. I got the case and I'm working on a fix for that. Thank you!

aaron-steinfeld: I believe so, although difficult to put into words. More abstractly, any filter tree must be convertible to `<subsetOfviewFilter> && <restOfFilterTree>`.
That could happen in a two different ways, it's a fun tree problem. Recursing from root, a node matches if:
- It's AND and at least one child matches
- It's OR and every child matches

That way:
- Match: `(FOO = x AND BAR = y) OR (FOO = x AND BAR = z)
- No Match: `(FOO = x OR BAR = y) AND (FOO = x OR BAZ = z)

Just for fun... I went down a logic rabbit hole when writing examples, so I guess the above doesn't always hold. Take
`(FOO = x OR BAR = y) AND (FOO = x OR BAR = z)
  ```
(FOO = x OR BAR = y) AND (FOO = x OR BAR = z) -> (un)distribute
(FOO = x ) OR (BAR = y AND BAR = z) -> (mutually exclusive)
(FOO = x ) OR FALSE -> (identity)
FOO = x
  ```
 Not sure how deep we want to go with our boolean logic...

aaron-steinfeld: Hah missed your second comment because I got carried away, but hopefully my comment is useful, else it was fun to write.

aaron-steinfeld:
Also just realized this will have some implications on attribute projections, as projections will need to be resolved before we do view filtering. I guess I'll do a full review.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
aaron-steinfeld: It can't - protos are never null!
```java
  public org.hypertrace.core.query.service.api.ColumnIdentifier getColumnIdentifier() {
    if (valueCase_ == 1) {
       return (org.hypertrace.core.query.service.api.ColumnIdentifier) value_;
    }
    return org.hypertrace.core.query.service.api.ColumnIdentifier.getDefaultInstance();
  }
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -155,6 +317,13 @@ public void handleRequest(
       RequestAnalyzer requestAnalyzer) {
     long start = System.currentTimeMillis();
     validateQueryRequest(queryContext, request);
+
+    // Remove the filters which match the view column filters because they shouldn't be passed down
```
aaron-steinfeld, line 321: That's not strictly true. Take the case of a view filtered to `FOO IN (X, Y)` and a filter provided that matches the view, `FOO = X` - it needs to be rewritten to `FOO = Y` (or `FOO IN (Y)`) when calling pinot. We should either handle that case, or remove the subset match logic and only take exact matches.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewColumnFilter.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public ViewColumnFilter(Operator operator, Set<String> values) {
+    this.operator = operator;
+    this.values = values;
+  }
+
+  public static ViewColumnFilter from(Config filterConfig) {
+    Operator operator = Operator.valueOf(filterConfig.getString("operator"));
+    return new ViewColumnFilter(operator, new HashSet<>(filterConfig.getStringList("values")));
```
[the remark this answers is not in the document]
buchi-busireddy: Yeah I was debating if I should use `value` for `EQ` and `values` for `IN` but I thought that might make it complex. I can fix it if you think that's fine, I don't have a strong feeling.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewDefinition.java
```diff
@@ -70,7 +74,15 @@ public static ViewDefinition parse(Config config, String tenantColumnName) {
       }
       columnSpecMap.put(logicalName, spec);
     }
-    return new ViewDefinition(viewName, columnSpecMap, tenantColumnName);
+
+    final Map<String, ViewColumnFilter> filterMap = new HashMap<>();
+    if (config.hasPath("filters")) {
+      for (Config filterConfig: config.getConfigList("filters")) {
```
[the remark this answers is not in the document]
buchi-busireddy: I'll add comments.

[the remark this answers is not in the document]
buchi-busireddy: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,6 +185,130 @@ public QueryCost canHandle(
...
+  private boolean doesFiltersMatch(ViewColumnFilter viewColumnFilter, Filter queryFilter) {
+    if (viewColumnFilter.getOperator() == ViewColumnFilter.Operator.IN) {
+      if (queryFilter.getOperator() != Operator.IN && queryFilter.getOperator() != Operator.EQ) {
+        return false;
+      }
+
+      return isSubSet(viewColumnFilter.getValues(), queryFilter.getRhs());
+    } else {
+      // query filter must be an equals filter in this case.
+      if (queryFilter.getOperator() != Operator.EQ) {
```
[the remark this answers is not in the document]
buchi-busireddy: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,6 +185,130 @@ public QueryCost canHandle(
...
+    }
+
+    Queue<Filter> filters = new ArrayDeque<>();
+    filters.add(request.getFilter());
+    Map<String, Filter> filterMap = new HashMap<>();
+
+    while (!filters.isEmpty()) {
+      Filter filter = filters.poll();
+
+      // Check if the filter is a leaf filter or not.
```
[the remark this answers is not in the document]
buchi-busireddy: Agree. I also realized a bit late. I think we can fix this, though in a different PR. Definitely fixing this will make it cleaner.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,6 +185,130 @@ public QueryCost canHandle(
...
+      if (filter.getChildFilterCount() > 0) {
+        filters.addAll(filter.getChildFilterList());
+      } else {
+        filterMap.put(filter.getLhs().getColumnIdentifier().getColumnName(), filter);
```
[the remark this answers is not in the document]
buchi-busireddy: Fixed this now.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewColumnFilter.java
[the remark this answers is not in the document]
aaron-steinfeld: I'd probably support either key in the config, reading as string for the single and list for the plural. You could then put either into the set and internally treat them as the same - that way we don't sacrifice the API for implementation complexity.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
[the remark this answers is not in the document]
aaron-steinfeld: yep, certainly separate from this PR.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,63 +148,110 @@ public QueryCost canHandle(
...
+
+    Map<String, Filter> childLeafFilterMap = filter.getChildFilterList().stream()
+        .filter(f -> f.getChildFilterCount() == 0 &&
+            f.getLhs().getValueCase() == Expression.ValueCase.COLUMNIDENTIFIER)
+        .filter(
+            f -> viewFilterMap.containsKey(f.getLhs().getColumnIdentifier().getColumnName()))
+        .collect(Collectors
+            .toMap(f -> f.getLhs().getColumnIdentifier().getColumnName(), f -> f));
+    if (childLeafFilterMap.isEmpty()) {
+      long matchCount = filter.getChildFilterList().stream()
```
aaron-steinfeld, line 210: If we flip the operator check, we can make this clearer by not using a match count but instead explicit predicates:

```
if (filter.getOperator() == Operator.AND) {
  return filter.getChildFilterList().stream().anyMatch(f -> doViewFiltersMatch(f, viewFilterMap));
} else if (filterOperator() === Operator.OR) {
  return filter.getChildFilterList().stream().allMatch(f -> doViewFiltersMatch(f, viewFilterMap));
}
```
(alternatively, drop the map above and use that in

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,63 +148,110 @@ public QueryCost canHandle(
...
+          .filter(b -> b).count();
+      if (filter.getOperator() == Operator.AND && matchCount > 0) {
+        return true;
+      }
+
+      // When children are being OR'ed, all the children must match the view filters.
+      return filter.getOperator() == Operator.OR && matchCount == filter.getChildFilterCount();
+    }
+
+    if (filter.getOperator() != Operator.AND) return false;
```
aaron-steinfeld, line 221: missing brackets. Also is this the case now where we have at least one leaf child? Why can't that match in the OR case? Admittedly, it's probably less common.
```
foo = x OR (foo = x AND bar = y) // Matches foo = x
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -147,63 +148,110 @@ public QueryCost canHandle(
...
+
+    for (Map.Entry<String, ViewColumnFilter> entry : viewFilterMap.entrySet()) {
```
aaron-steinfeld, line 223: We seem to go back and forth between for and stream quite a bit. This also appears to error if we don't have a child filter that references one of the view filters (as 225 would return null), which I would expect to be a regular false condition. I think we avoid it with one of the short circuits, but it doesn't seem necessary as it just makes things more brittle.
As a stream:
```
return viewFilterMap.entrySet().stream()
   .allMatch(viewFilterEntry -> 
childLeafFilterMap.containsKey(entry.getKey()) && doesSingleViewFilterMatchLeafQueryFilter(entry.getValue(), childLeafFilterMap.get(entry.getKey())));
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Implemented this and handling most (all?) of the cases that we discussed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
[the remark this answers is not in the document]
buchi-busireddy: Logged https://traceableai.atlassian.net/browse/ENG-6061 to take care of this

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Fixed this and added a test case to cover this case.

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryRequestBuilderUtils.java
```diff
@@ -38,18 +39,31 @@ public static Filter createEqualsFilter(String column, String value) {
...
+
   public static Filter createFilter(String column, Operator operator, Expression expression) {
     return Filter.newBuilder().setLhs(createColumnExpression(column))
         .setOperator(operator).setRhs(expression).build();
   }
 
   public static Expression createStringLiteralValueExpression(String value) {
     return Expression.newBuilder().setLiteral(LiteralConstant.newBuilder().setValue(
-        Value.newBuilder().setString(value))).build();
+        Value.newBuilder().setString(value).setValueType(ValueType.STRING))).build();
```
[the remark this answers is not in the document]
buchi-busireddy: This was written long time ago and I don't remember if there was any reason. It's just that we got started there and continued that way AFAIK.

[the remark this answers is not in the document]
aaron-steinfeld: Looking at value and trying to reach way back into the mists of the past, I believe it was because oneof doesn't support repeated and maybe map fields. A wrapper would also have sufficed there, but that was likely the reason.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewColumnFilter.java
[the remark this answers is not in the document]
buchi-busireddy: Addressed.

buchi-busireddy:
@tim-mwangi @aaron-steinfeld Addressed all the comments or otherwise replied to them. Added unit tests for different scenarios. This is ready for review again.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -171,61 +168,42 @@ public QueryCost canHandle(QueryRequest request, Set<String> referencedSources,
...
+            .filter(b -> b).count();
 
-      for (Map.Entry<String, ViewColumnFilter> entry : viewFilterMap.entrySet()) {
-        // Make sure all view filters match.
-        Filter queryFilter = childLeafFilterMap.get(entry.getKey());
-        if (queryFilter == null || !doesSingleViewFilterMatchLeafQueryFilter(entry.getValue(), queryFilter)) {
-          // Few columns matched but others didn't match.
-          return false;
-        }
+        return filter.getOperator() == Operator.AND && matchCount == viewFilterMap.size();
```
aaron-steinfeld, line 205: This is checking for a "full match" but misses a number of scenarios. All we know about these children is that at least one is a leaf, but not necessarily all.
In the AND case, we need each view filter to be matched, but that could happen in one or more nodes. It's also not necessarily indicated by the count as we could have duplicates:

Duplicate case:
View Filter: X = 'A' AND Y = 'B'
Query Filter: X = 'A' AND X= 'A'

Single match case (this specific example wouldn't happen given our optimize logic, but are we intending on relying on that here? that seems like a far distance for an assumption to travel):
View Filter: X = 'A' AND Y = 'B'
Query Filter: (X = 'A' AND Y= 'B') AND Z = 'C'

In the OR case, we need each child to be a full match, but right now we'd never match since we only look at ANDs

Each full match case:
View Filter: X = 'A'
Query Filter: X = 'A' OR X = 'A' // many equivalencies to this

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewColumnFilter.java
```diff
@@ -26,7 +39,11 @@ public ViewColumnFilter(Operator operator, Set<String> values) {
 
   public static ViewColumnFilter from(Config filterConfig) {
     Operator operator = Operator.valueOf(filterConfig.getString("operator"));
-    return new ViewColumnFilter(operator, new HashSet<>(filterConfig.getStringList("values")));
+    if (operator == Operator.EQ) {
+      return new ViewColumnFilter(operator, Set.of(filterConfig.getString("value")));
+    } else {
+      return new ViewColumnFilter(operator, new HashSet<>(filterConfig.getStringList("values")));
```
aaron-steinfeld, line 45: `Set.copyOf` to match 43, and keep things immutable

buchi-busireddy: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Yes, I'm relying on the filter optimization for all these 3 cases and other similar duplicate cases or flattening predicates. I feel that keeps the tree navigation logic much simpler instead of handling so many cases.
Reg keeping the filter optimization far from this code: I thought about it and the reason I'm doing it early on is 
1. so that we don't have to repeat it in each handler
2. other parts of the code like conversion to SQL can take advantage of it too.

The optimization logic is in a separate class and can be invoked from here too if needed but I think this is fine.

aaron-steinfeld: See the later part of my most recent comment - https://github.com/hypertrace/query-service/pull/17#discussion_r489412091  - I believe this proposes a way of doing this in a much more straightforward way that doesn't require (but works with) optimization and all scenarios.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -171,61 +168,42 @@ public QueryCost canHandle(QueryRequest request, Set<String> referencedSources,
...
+   * viewFilterMap.
+   *
+   * @return True if the given filter node matches all the view filters from the given map, False
+   * otherwise.
+   */
   private boolean doViewFiltersMatch(Filter filter, Map<String, ViewColumnFilter> viewFilterMap) {
-    // Navigate the entire filter tree to check if every query filter which has the
-    // view filter columns has all matching view filters. If not, return false since
-    // this view can't handle such query.
+    // 1. Basic case: Filter is a leaf node. This will be a match only if there is only one
```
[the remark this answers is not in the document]
buchi-busireddy: Case 3 handles the example you have given here. In last iteration I was checking for viewFilter count but I found this logic to be simpler and cleaner.

The approach doesn't support partial match. Are there any pieces of code which seem to indicate that? I can add comments/clarify if needed but the whole logic only checks for full matches. Updating the javadoc of different classes and methods to make this clear.

Reg the other approach you're suggesting, the v1 code I've written was on those same lines but what complicated the solution is, there could be multiple matching filter nodes in the tree because of duplicate predicates. So, we should be able to return multiple matching filter nodes. I can take another fresh look at that approach in another PR but I think this can go because it's functionally correct and code is easy enough.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -171,61 +168,42 @@ public QueryCost canHandle(QueryRequest request, Set<String> referencedSources,
...
-      }
-
-      if (filter.getOperator() != Operator.AND)
-        return false;
+        return (filter.getOperator() == Operator.AND && matchCount > 0) ||
+            (filter.getOperator() == Operator.OR && matchCount == filter.getChildFilterCount());
+      } else {
+        // 3. An internal filter with children which are leaves: All the view filters
+        // should be matched and AND'ed, otherwise it's not a match.
+        long matchCount = filter.getChildFilterList().stream()
```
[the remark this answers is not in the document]
buchi-busireddy: Fixed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -171,61 +168,42 @@ public QueryCost canHandle(QueryRequest request, Set<String> referencedSources,
...
+          .filter(f -> f.getChildFilterCount() == 0)
           .collect(Collectors
               .toMap(f -> f.getLhs().getColumnIdentifier().getColumnName(), f -> f));
+
+      // 2. An internal filter node which doesn't have any leaves but other internal nodes:
+      // Parent is a match if all the OR'ed children match or at least one of AND'ed children match.
       if (childLeafFilterMap.isEmpty()) {
         long matchCount = filter.getChildFilterList().stream()
             .map(f -> doViewFiltersMatch(f, viewFilterMap))
             .filter(b -> b).count();
```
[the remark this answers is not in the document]
buchi-busireddy: Fixed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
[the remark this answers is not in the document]
aaron-steinfeld: > Case 3 handles the example you have given here. In last iteration I was checking for viewFilter count but I found this logic to be simpler and cleaner.

Maybe I'm misreading this, but given a single leaf filter regardless of how many view filters we have, we'd return true on 192 and go into the singleViewMatch function. That function would only check if the leaf filter matches the corresponding col view filter - in other words, a partial match. Later, we rely on this returning true for a partial match, when we do things like get the match count and make sure it matches the number of view cols (line 212).

> Reg the other approach you're suggesting, the v1 code I've written was on those same lines but what complicated the solution is, there could be multiple matching filter nodes in the tree because of duplicate predicates. So, we should be able to return multiple matching filter nodes. I can take another fresh look at that approach in another PR but I think this can go because it's functionally correct and code is easy enough.

Not sure I follow the issue here - if we're maintaining a set of matches, duplicates wouldn't matter. I don't believe this is functionally correct, let's talk it through.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -189,6 +369,58 @@ public void handleRequest(
...
+  @Nonnull
+  private QueryRequest rewriteRequestWithViewFiltersApplied(QueryRequest request,
+      Map<String, ViewColumnFilter> columnFilterMap) {
+
+    Filter newFilter = removeViewColumnFilter(request.getFilter(), columnFilterMap);
+    QueryRequest.Builder builder = QueryRequest.newBuilder(request).clearFilter();
+    if (!Filter.getDefaultInstance().equals(newFilter)) {
+      builder.setFilter(newFilter);
+    }
+    return builder.build();
```
aaron-steinfeld, line 381: Query should work just fine with a default empty filter (as that's what getFilter will always return anyway, never null)
```suggestion
    return QueryRequest.newBuilder(request).setFilter(newFilter).build();
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -189,6 +369,58 @@ public void handleRequest(
...
+  }
+
+  private Filter removeViewColumnFilter(Filter filter,
+      Map<String, ViewColumnFilter> columnFilterMap) {
+    if (filter.getChildFilterCount() > 0) {
+      // Recursively try to remove the filter and eliminate the null nodes.
+      Set<Filter> newFilters = filter.getChildFilterList().stream()
+          .map(f -> removeViewColumnFilter(f, columnFilterMap))
+          .filter(f -> !Filter.getDefaultInstance().equals(f))
+          .collect(Collectors.toCollection(LinkedHashSet::new));
```
aaron-steinfeld, line 391: nit: if we don't have a particular need for a linked hash set, there's shortcuts like `Collectors.toSet()` or `Collectors.toUnmodifiableSet()`

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -189,6 +369,58 @@ public void handleRequest(
...
+
+      if (newFilters.isEmpty()) {
+        return Filter.getDefaultInstance();
+      } else if (newFilters.size() == 1) {
+        return newFilters.stream().findFirst().get();
```
aaron-steinfeld, line 396: nit - believe we're already using guava in here, so a little clearer to do:
```suggestion
        return Iterables.getOnlyElement(arguments);
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -189,6 +369,58 @@ public void handleRequest(
...
+
+  private Filter rewriteLeafFilter(Filter queryFilter,
+      Map<String, ViewColumnFilter> columnFilterMap) {
+    ViewColumnFilter viewColumnFilter =
+        columnFilterMap.get(queryFilter.getLhs().getColumnIdentifier().getColumnName());
+    if (viewColumnFilter == null) {
+      return queryFilter;
+    } else {
+
+      // The only case where we need to return non-null filter is when view filter is based on 'IN'
```
aaron-steinfeld, line 413: are there cases here with different RHS? like:
VF: `X = 'A'`
QF: `X = 'A' AND X ='B'` - should return nothing

VF: `X IN ('A', 'B')`
QF: `X IN ('A', 'B') AND X ='B'` - should return only B matches

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -186,31 +177,26 @@ public QueryCost canHandle(QueryRequest request, Set<String> referencedSources,
    * @return True if the given filter node matches all the view filters from the given map, False
    * otherwise.
    */
-  private boolean doViewFiltersMatch(Filter filter, Map<String, ViewColumnFilter> viewFilterMap) {
-    // 1. Basic case: Filter is a leaf node. This will be a match only if there is only one
-    // view filter and that matches the leaf filter.
+  private Set<String> getMatchingViewFilterColumns(Filter filter, Map<String,
```
aaron-steinfeld, line 177: beautiful!

aaron-steinfeld (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
aaron-steinfeld: @buchi-busireddy - this is the last comment I think we need to make sure we resolve (either through comment or fix)

aaron-steinfeld (changes requested):
Got excited and approved with the rewrite, but there's one important unresolved comment so removing (can't figure out how to just remove approval)

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Thanks for the thorough review and ideas.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: I wanted to retain the order, mainly because I was verifying sql query in the unit test and without order the predicates were sometimes coming up in different order.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: Oh this is cool. Will fix it.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
buchi-busireddy: You need to get 5 ⭐ for your reviews 👍 

I've fixed this case and added unit test also now.

aaron-steinfeld (approved):

tim-mwangi (approved):
I'm gonna assume this was rigorously reviewed 😂

PR #21, opened 2020-09-18
feat: add pinot function registry to translate functions

aaron-steinfeld:


query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryFunctionConstants.java
```diff
@@ -0,0 +1,15 @@
+package org.hypertrace.core.query.service;
+
+import org.hypertrace.core.query.service.api.Expression;
+import org.hypertrace.core.query.service.api.Function;
+
+/** A converter from the canonical format received in a query to the appropriate dat */
```
tim-mwangi, line 6: typo: "dat"?

tim-mwangi (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryFunctionConstants.java
aaron-steinfeld: It looks like I got distracted in the midst of writing this comment... And then I changed this from an implemented interface to constants.

```suggestion
/**
 * Canonical query function names received in requests. These should be converted to data-store
 * specific names when handling a request.
 */
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryFunctionConstants.java
```diff
@@ -0,0 +1,15 @@
+package org.hypertrace.core.query.service;
+
+import org.hypertrace.core.query.service.api.Expression;
+import org.hypertrace.core.query.service.api.Function;
```
aaron-steinfeld, line 4:
```suggestion
```

tim-mwangi (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -0,0 +1,130 @@
...
+
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.hypertrace.core.query.service.api.Expression;
+import org.hypertrace.core.query.service.api.Function;
+import org.hypertrace.core.query.service.api.LiteralConstant;
+import org.hypertrace.core.query.service.api.Value;
+import org.hypertrace.core.query.service.api.ValueType;
+
+public class PinotFunctionConverter {
```
buchi-busireddy, line 14: Nice.. Thanks for cleaning this up!

aaron-steinfeld: No problem. Once this is up and running, we can do some clean up on the gateway side to strip the conversion there since this handles it now.

aaron-steinfeld:
Change Motivation: Query service is an abstraction over one or more data stores, but currently the abstract query requires implementation-specific function names. This function converter relaxes that constraint, and converts abstract functions (which, as of this change, are mostly the same as pinot-specific functions, with the exception of percentile and count) into functions compatible with pinot. This will allow us to remove some pinot-bound logic in gateway service (which previously did this conversion - we include logic here to unwind that for backwards compatibility), as well as add an upcoming feature, attribute projection, in such a way that the projection logic is not coupled to a specific data store.

PR #23, opened 2020-09-23
refactor: update config for immutability and extension

aaron-steinfeld:
This change is to support adding further configuration in the form of attribute client config. When adding this, the use of ConfigBeanFactory prevented our standard dotted notation (because each top level property must match a property in the pojo). Digging in further, there were a number of antipatterns with the bean factory and existing config parsing- for example, the coupling of method names and config, the distribution of parsing logic across the codebase, the mutability of configs and confusing naming.

This change takes a cut at cleaning this up in code, while leaving the config format intact to ensure compatibility with existing configs.


query-service-impl/src/test/resources/application.conf
```diff
@@ -2,6 +2,12 @@ service.name = "query-service"
 service.port = 8090
 service.admin.port = 8091
 service.config = {
+  attribute.client = {
```
buchi-busireddy, line 5: Where are you using these? I don't see the code to validate attributes yet.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final String CONFIG_PATH_CLIENT_KEY = "clientConfig";
+    private static final String CONFIG_PATH_REQUEST_HANDLER_INFO = "requestHandlerInfo";
+
+    private final String name;
+    private final String type;
+    private final String clientConfig;
+    private final Config requestHandlerInfo;
+
+    private RequestHandlerConfig(Config config) {
+      this.name = config.getString(CONFIG_PATH_NAME);
```
buchi-busireddy, line 55: Just for my own understanding, what's the difference b/w this model and the previous one like:
`ConfigBeanFactory.create(config, QueryServiceImplConfig.class);`?

buchi-busireddy (approved):

query-service-impl/src/test/resources/application.conf
aaron-steinfeld: These actually aren't used yet, they're used in the following change for projections - that change drove this one, so I split this part off to shrink the review. Should have probably removed these.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java
aaron-steinfeld: Discussed in the PR description - https://github.com/hypertrace/query-service/pull/23#issue-491929367 
Specifically, the problem that drove this was the fact that every top level property in the bean object had to correspond to a top level config property.

tim-mwangi (approved):

PR #24, opened 2020-09-24
refactor: deprecate query utils in api, move any used to impl

aaron-steinfeld:
utils do not belong in an api package. deprecated them for backwards
compatibility and removed all usages from query service code. Moved
those used into a file inside the impl code, and those used only by
tests to the exist test util.


query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryRequestUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import org.hypertrace.core.query.service.api.Value;
+import org.hypertrace.core.query.service.api.ValueType;
+
+/**
+ * Utility methods to easily create {@link org.hypertrace.core.query.service.api.QueryRequest} its
+ * selections and filters.
+ */
+public class QueryRequestUtil {
+
+  public static Expression createColumnExpression(String columnName) {
```
aaron-steinfeld, line 15: Not too worried about test coverage on these guys - by their nature, to verify them, we'd be effectively duplicating the code in the test anyway.

query-service-api/src/main/java/org/hypertrace/core/query/service/util/QueryRequestUtil.java
```diff
@@ -129,23 +130,17 @@ public static Filter createTimeFilter(String columnName, Operator op, long value
             Expression.newBuilder()
                 .setLiteral(
                     LiteralConstant.newBuilder()
-                        .setValue(
-                            org.hypertrace.core.query.service.api.Value.newBuilder()
-                                .setString("1:MILLISECONDS:EPOCH"))))
+                        .setValue(Value.newBuilder().setString("1:MILLISECONDS:EPOCH"))))
```
tim-mwangi, line 133: All nice and good until you realize this is used in gateway-service.

aaron-steinfeld: Yeah, I completely removed it in my original change and then found that out the fun way when testing locally. Hence the change to deprecation here.

tim-mwangi (approved):

PR #25, opened 2020-09-25
feat: add support for projected attributes

aaron-steinfeld:
This change adds support for pushing down attribute projections sourced from an attribute definition into an equivalent query expression. Ultimately, this will allow defining new attributes at runtime without having to make any code or storage changes, as long as they can be expressed in our limited (to be expanded) projection language in terms of existing attributes.


query-service-impl/src/main/java/org/hypertrace/core/query/service/AttributeClientProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory.getClientCallCredsProvider;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import javax.inject.Inject;
+import javax.inject.Provider;
+import org.hypertrace.core.attribute.service.cachingclient.CachingAttributeClient;
+
+class AttributeClientProvider implements Provider<CachingAttributeClient> {
```
jcchavezs, line 11: nit, can/should this be final?

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -0,0 +1,282 @@
...
+import org.hypertrace.core.query.service.api.ColumnIdentifier;
+import org.hypertrace.core.query.service.api.Expression;
+import org.hypertrace.core.query.service.api.Filter;
+import org.hypertrace.core.query.service.api.Function;
+import org.hypertrace.core.query.service.api.OrderByExpression;
+import org.hypertrace.core.query.service.api.QueryRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class ProjectionTransformation implements QueryTransformation {
```
jcchavezs, line 33: same nit about final.

query-service-impl/src/main/java/org/hypertrace/core/query/service/AttributeClientProvider.java
aaron-steinfeld: No reason - I can't imagine a scenario where this would be extended. Unlike variables, I don't generally mark classes as final by default, but I don't really have any justification for it since the same reasoning for variables applies to classes, if not more so (preventing the inheritance scourge!).

Thanks for pointing this out, will update these two spots and try to get into the habit of doing this going forward.

aaron-steinfeld: Thinking this through a bit more - I guess, as you had already pointed out, this really applies for implementations that shouldn't be referenced directly. A big reason not to is it makes things difficult to mock, but in these scenarios (and any where the implementation shouldn't be referenced directly) the interface gives you an adequate and more appropriate handle to mock anyway.

tim-mwangi: I was gonna say final classes cannot be mocked but @aaron-steinfeld brought up a good point about using interfaces instead of direct impls.

query-service-impl/src/main/java/org/hypertrace/core/query/service/AttributeClientProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory.getClientCallCredsProvider;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import javax.inject.Inject;
+import javax.inject.Provider;
+import org.hypertrace.core.attribute.service.cachingclient.CachingAttributeClient;
+
+final class AttributeClientProvider implements Provider<CachingAttributeClient> {
```
tim-mwangi, line 11: This is fun!

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -0,0 +1,282 @@
...
+          transformed);
+    }
+  }
+
+  /**
+   * This doesn't change any functional behavior, but omits fields that aren't needed, shrinking the
+   * object and keeping it equivalent to the source object for equality checks.
+   */
+  private Filter rebuildFilterOmittingDefaults(
+      Filter original, Expression lhs, Expression rhs, List<Filter> chidFilters) {
```
tim-mwangi, line 238: typo: childFilters

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -0,0 +1,282 @@
...
+  private Single<Expression> transformExpression(Expression expression) {
+    switch (expression.getValueCase()) {
+      case COLUMNIDENTIFIER:
+        return this.transformColumnIdentifier(expression.getColumnIdentifier());
+      case FUNCTION:
+        return this.transformFunction(expression.getFunction())
+            .map(expression.toBuilder()::setFunction)
+            .map(Expression.Builder::build);
+      case ORDERBY:
+        return this.transformOrderBy(expression.getOrderBy())
```
tim-mwangi, line 74: 74-76 not unit tested

aaron-steinfeld: So I actually meant to ask about this. I'm unclear if this is a missing test, or if this an issue in the modeling? In what scenario is an expression of type orderBy?  Order by is set at the top level of the query as an expression, which can be a function or column identifier. The request from the order by test looks like:

```java
    QueryRequest originalRequest =
        QueryRequest.newBuilder()
            .addSelection(createColumnExpression(PROJECTED_ATTRIBUTE_ID))
            .addOrderBy(createOrderByExpression(PROJECTED_ATTRIBUTE_ID, SortOrder.DESC))
            .build();
```
where the order by expression is built by:
```java
  public static OrderByExpression.Builder createOrderByExpression(
      String columnName, SortOrder order) {
    return OrderByExpression.newBuilder()
        .setOrder(order)
        .setExpression(createColumnExpression(columnName));
  }
```

tim-mwangi: I see what you are saying. Seems like a problem with the API def.

tim-mwangi (approved):

PR #27, opened 2020-10-04
fix: use concatmap instead of flatmap to prevent race

aaron-steinfeld:


query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -58,7 +58,7 @@
 
   private Single<List<Expression>> transformExpressionList(List<Expression> expressionList) {
     return Observable.fromIterable(expressionList)
-        .flatMapSingle(this::transformExpression)
+        .concatMapSingle(this::transformExpression)
```
buchi-busireddy, line 61: Would it be sequential instead of parallel now? Any impact on the speed or parallellism?

buchi-busireddy (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -169,7 +169,7 @@
       ProjectionExpression projectionExpression) {
     Single<List<Expression>> argumentListSingle =
         Observable.fromIterable(projectionExpression.getArgumentsList())
-            .flatMapSingle(this::rewriteProjectionAsQueryExpression)
```
buchi-busireddy, line 172: @aaron-steinfeld Don't you need to change on L133 too?

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
aaron-steinfeld: Yes this makes it sequential. Should be negligible impact on speed - the only things that are async in this pipeline (attributes) are fetched together and cached so don't benefit from parallelism.

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
aaron-steinfeld: Line 133 is different (hence the initial mistake). When going from a maybe -> anything or single -> anything there's no source ordering to concern, so flatmap and concatmap are equivalent (concat's impl actually just calls flat). Going from something with its own order (an observable, or flowable which we aren't using), then you have to be careful when picking flat (merge), concat (sequential), or switch (latest).

PR #30, opened 2020-10-23
Fixing a bug in converting the filter with IN clause when column type is BYTES in Pinot.

buchi-busireddy:
Tested with Hypertrace helm setup.


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
```diff
@@ -15,21 +16,21 @@
   private final Map<Integer, String> stringParams;
   private final Map<Integer, Float> floatParams;
   private final Map<Integer, Double> doubleParams;
-  private final Map<Integer, String> bytesStringParams;
+  private final Map<Integer, ByteString> byteStringParams;
```
buchi-busireddy, line 19: Leaving the bytes params as bytes till the last minute will let us to do the correct transformation at the end.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -200,43 +199,18 @@ private Expression handleValueConversionForLiteralExpression(Expression lhs, Exp
       return rhs;
     }
 
-    Expression newRhs = rhs;
     String lhsColumnName = lhs.getColumnIdentifier().getColumnName();
-
-    ToValueConverter converter = getValueConverter(rhs.getLiteral().getValue().getValueType());
```
buchi-busireddy, line 206: @kotharironak I know you have added this. The requirement I see here is mainly around doing specific transformation on the destination column type in Pinot, not about what was the type in the incoming request. Hence, I thought we don't really need an interface since we'll always have only one implementation.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
```diff
@@ -15,21 +16,21 @@
...
       Map<Integer, Double> doubleParams,
-      Map<Integer, String> bytesStringParams) {
+      Map<Integer, ByteString> bytesParams) {
     this.integerParams = integerParams;
     this.longParams = longParams;
     this.stringParams = stringParams;
     this.floatParams = floatParams;
     this.doubleParams = doubleParams;
-    this.bytesStringParams = bytesStringParams;
+    this.byteStringParams = bytesParams;
```
kotharironak, line 33:
```suggestion
    this.byteStringParams = byteParams;
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
```diff
@@ -15,21 +16,21 @@
...
+  private final Map<Integer, ByteString> byteStringParams;
 
   private Params(
       Map<Integer, Integer> integerParams,
       Map<Integer, Long> longParams,
       Map<Integer, String> stringParams,
       Map<Integer, Float> floatParams,
       Map<Integer, Double> doubleParams,
-      Map<Integer, String> bytesStringParams) {
+      Map<Integer, ByteString> bytesParams) {
```
kotharironak, line 27:
```suggestion
      Map<Integer, ByteString> byteParams) {
```

tim-mwangi (dismissed):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/DestinationColumnValueConverter.java
```diff
@@ -0,0 +1,63 @@
...
+        for (String s : value.getStringArrayList()) {
+          valueBuilder.addBytesArray(convertToByteString(s));
+        }
+        valueBuilder.setValueType(ValueType.BYTES_ARRAY);
+        return valueBuilder.build();
+
+      default:
+        String msg = String.format("Unsupported value of type: %s while transforming to BYTES.",
+            value.getValueType().name());
+        LOG.warn(msg);
```
kotharironak, line 52: Do we need a `warn` message here as we are already throwing an exception that will be logged later, right?

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -685,35 +687,31 @@ public void testQueryWithBytesColumnWithEmptyId() {
...
     assertPQLQuery(
         builder.build(),
-        "SELECT span_id FROM SpanEventView "
+        "SELECT duration_millis FROM SpanEventView "
             + "WHERE "
             + viewDefinition.getTenantIdColumn()
             + " = '"
             + TENANT_ID
             + "' "
-            + "AND REGEXP_LIKE(parent_span_id,'')");
```
kotharironak, line 716: Don't we need a REGEX_LIKE operator as well? Is it blocked at the query/getaway layer as we still calling `handleValueConversionForLiteralExpression`  from the LIKE clause in `convertFilter2String`?

kotharironak (commented):
lgtm

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -747,37 +745,7 @@ public void testQueryWithStringColumnWithNullString() {
             + " = '"
             + TENANT_ID
             + "' "
-            + "AND ( span_id != 'null' ) limit 5");
-  }
-
-  @Test
-  public void testQueryWithStringColumnWithNullStringWithLikeFilter() {
```
buchi-busireddy, line 754: Removed this because REGEXP_LIKE doesn't make sense for `null`

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
buchi-busireddy: I've removed this because `REGEXP_LIKE` doesn't make sense for null or empty string. `col = ''` is the way to check for BYTES columns.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/DestinationColumnValueConverter.java
buchi-busireddy: I don't think we can make assumptions about the caller logging or not logging this. If this is useful, we should log here I think.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
buchi-busireddy: It's indeed bytes, not a single byte. Calling it `byteStringParams` might be even clearer.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
buchi-busireddy: It's indeed bytes, not a single byte. Calling it byteStringParams might be even clearer.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
kotharironak: sounds good

kotharironak (approved):

tim-mwangi (approved):

PR #33, opened 2020-10-30
fix: update pinot conversion for concat operator

aaron-steinfeld:


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -76,6 +80,14 @@ private Function toPinotPercentile(Function function) {
         .build();
   }
 
+  private Function toPinotConcat(Function function) {
+    // Our concat takes two args (one can be literal - CONCAT(ARG1, ARG2)
```
buchi-busireddy, line 84: Should we support the SEPARATOR in our CONCAT too? it's pretty useful.

tim-mwangi (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
aaron-steinfeld: So my feeling is that our concat is more representative of concat in most languages, and can already model this like `CONCAT(ARG1, CONCAT(SEP, ARG2))` (just like theirs can model ours like `CONCAT(ARG1, ARG2, "")`. If we do like the terser syntax, my preference would be to introduce an operator like `JOIN` which more traditionally takes a separator.

PR #40, opened 2020-11-24
feat: add support for typed null expressions

aaron-steinfeld:
## Description
This change adds support for explicitly passing null in an abstract query, without having to rely on the implementation's (i.e. pinot) conception of null. This is then used to correctly convert projection expressions containing null based on their target data type.

https://github.com/hypertrace/hypertrace/issues/126

### Testing
Added unit tests for associated functionality, also tested the full HT cluster with a nullable attribute locally 

### Checklist:


query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -159,17 +174,49 @@
...
+        return Single.error(
+            new UnsupportedOperationException(
+                "Unrecognized attribute kind for null expression conversion: " + attributeKind));
+    }
+  }
+
   private Single<Expression> rewriteProjectionExpressionAsQueryExpression(
       ProjectionExpression projectionExpression) {
+    List<Projection> arguments = projectionExpression.getArgumentsList();
+    // TODO fallback for unknown projections
```
tim-mwangi, line 211: fixing this one?

aaron-steinfeld: good thing I put a todo in, forgot to come back to it :) will update.

tim-mwangi: good thing I had taken off my trust cap :)

aaron-steinfeld: Updated - I wrote a test for it, but it's actually not testable since we cover all known projections in code and its an enum. I modified the code to not cover one case and ran the test :) Any other ideas for testing the situation where it's an unknown enum value?

tim-mwangi: Nope. No idea lol. Was gonna say you could pass in a int value but compiler will not let you. And if you some reflection magic, then the reflection fails when looking for the enum to convert the int to.

tim-mwangi (approved):

PR #42, opened 2020-11-25
chore: update gradle

aaron-steinfeld:
## Description
Update gradle to latest version

### Testing
Built locally

### Checklist:
- [n/a] I have added tests that prove my fix is effective or that my feature works


JBAhire (approved):

query-service/build.gradle.kts
```diff
@@ -24,7 +24,7 @@ dependencies {
 }
 
 application {
-  mainClassName = "org.hypertrace.core.serviceframework.PlatformServiceLauncher"
+  mainClass.set("org.hypertrace.core.serviceframework.PlatformServiceLauncher")
```
tim-mwangi, line 27: interesting

tim-mwangi (approved):

query-service/build.gradle.kts
aaron-steinfeld: this was just a deprecation that I noticed, the old way would have also worked. In general, gradle is moving to use properties (which support lazy resolution) vs values. Most of our own plugins already use the property syntax.

PR #47, opened 2020-12-29
ci: migrating to GitHub actions

JBAhire:
## Description
Migrating to GitHub actions from CircleCI as per https://github.com/hypertrace/hypertrace/issues/144


.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,78 @@
...
+            gradle-packages-${{ runner.os }}-${{ github.job }}
+            gradle-packages-${{ runner.os }}
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_READ_USER }}
+          password: ${{ secrets.DOCKERHUB_READ_TOKEN }}
+      
+      - name: Build with Gradle
```
aaron-steinfeld, line 41: Before rolling out across all repos, could we address the two pieces of the existing CI config that didn't get ported:
- Reporting test results (already did this as part of the ui port - example here https://github.com/hypertrace/hypertrace-ui/commit/20072af17d37c0f92ea79dd9b212457e5e63d204#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948R85-R90)
- A gradle action that standardizes the CLI flags to correctly set parallelism, logging etc. for CI

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          fetch-depth: 0
+      
+      - name: create checksum file
+        uses: hypertrace/actions/checksum@main
+
+      - name: Cache packages
+        id: cache-packages
```
aaron-steinfeld, line 26: nit: `id` is unused and thus unnecessary

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,78 @@
...
+          username: ${{ secrets.DOCKERHUB_READ_USER }}
+          password: ${{ secrets.DOCKERHUB_READ_TOKEN }}
+      
+      - name: Build with Gradle
+        run: ./gradlew build dockerBuildImages
+
+  validate-helm-charts:
+    runs-on: ubuntu-20.04
+    container: 
+      image: hypertrace/helm-gcs-packager:0.3.1
```
aaron-steinfeld, line 47: just a thought - could we move this container inside the our validate-charts + package-charts actions?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Check out code
+        uses: actions/checkout@v2.3.4
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          fetch-depth: 0
+      - name: Setup snyk
+        uses: snyk/actions/setup@0.3.0
+      - name: Snyk test
+        run: snyk test --all-sub-projects --org=hypertrace --severity-threshold=low --policy-path=.snyk --configuration-matching='^runtimeClasspath$'
```
aaron-steinfeld, line 76: also may be worth moving this into an action at some point

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Copy test reports
+        run: ./gradlew copyAllReports --output-dir=/tmp/test-reports
+
+      - name: Archive test reports
+        uses: actions/upload-artifact@v1
+        with:
+          name: test-reports
+          path: /tmp/test-reports
+        if: ${{ always() }}
```
aaron-steinfeld, line 49: I think these braces are unnecessary as `if` is always an expression

aaron-steinfeld (commented):
A few thoughts on optimizations before broader rollouts

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,49 @@
...
+        id: cache-packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle
+          key: gradle-packages-${{ runner.os }}-${{ github.job }}-${{ hashFiles('**/checksum.txt') }}
+          restore-keys: |
+            gradle-packages-${{ runner.os }}-${{ github.job }}
+            gradle-packages-${{ runner.os }}
+
+      - name: Unit test
```
kotharironak, line 31: There is no `integration` test in this repo?

kotharironak: Just checked, it's not there.

tim-mwangi: yeah, nope :)

.github/workflows/pr-build.yml
JBAhire: First one, I can add easily. Little confused on second one here. I explored a path to create gradle composite action based on arguments but composite actions don't support if conditions as of now so I dropped that idea.

.github/workflows/pr-build.yml
JBAhire: I think, we can't move this step to action as it's already using other action. let me try this one.

aaron-steinfeld: the setup itself would have to stay out (still crazy to me that GH doesn't support this yet, suspect it must be coming soon), but the test could be its own action right? or we could make the test run in a containerized action which means setup isn't needed

.github/workflows/pr-build.yml
JBAhire: We can but I have some doubts about it. check this out: https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-docker-container-action

.github/workflows/pr-build.yml
aaron-steinfeld: I was basically just thinking something kind of like below (not sure if this is quite right, but gives the idea)
```yaml
name: 'Gradle Command'
inputs:
  args:
    required: true
runs:
  using: "composite"
  steps: 
    - run: |
          ./gradlew ${{ inputs.args }} --info --max-workers=2 -Dorg.gradle.jvmargs=-Xmx2g -Dorg.gradle.console=plain --continue
      shell: bash
```

.github/workflows/pr-build.yml
aaron-steinfeld: What are the doubts? I used a docker containerized action as part of https://github.com/hypertrace/hypertrace-ui/commit/20072af17d37c0f92ea79dd9b212457e5e63d204#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948R85-R90 (you can use the action directly and let it build it each time, or just reference the container) and it's pretty easy on the consumer side. Not sure how much of a pain it is to produce/maintain.

JBAhire: will do this.

.github/workflows/pr-build.yml
JBAhire: for this, I would wait as they said they will bring support for uses and if in near future. We would like to abstract this at that time anyways. :)

Let me know if that will be fine. We can surely move test but that's a single command so I won't bother a lot about it.

.github/workflows/pr-build.yml
JBAhire: Okay. I can take a stab at this. seems straightforward. But if we are looking for gradle actions anyways can we also look at already available actions like: 
- https://github.com/marketplace/actions/gradle-command
- https://github.com/burrunan/gradle-cache-action

or we want to do something like above only?

aaron-steinfeld: So the other two tasks solve different problems - mainly (from my understanding after a quick looko) resolution of gradle itself and caching. I think ideally, we'd incorporate them into our gradle action too, but those aren't pain points we have right now - the more pressing need is an ability to propagate the cli flags we want for gradle without having to hardcode them at each invocation.

.github/workflows/pr-build.yml
aaron-steinfeld: Up to you. The problem I'm trying to avoid is that the argument set of this command has evolved over time, and now is mismatched in different repos depending on when their ci was set up. Capturing even the single line in an action avoids that (similar to our checksum action), but perhaps it's stable enough now that it's less of a concern and not worth the effort.

.github/workflows/pr-build.yml
JBAhire: I created gradle action and tested it here: https://github.com/hypertrace/attribute-service/pull/65

.github/workflows/pr-build.yml
JBAhire: also tested here: hypertrace/attribute-service#65

.github/workflows/pr-build.yml
JBAhire: reporting test result action is not able to parse this reports: https://github.com/hypertrace/attribute-service/runs/1642325282?check_suite_focus=true#step:10:1

JBAhire: @aaron-steinfeld , can you check once? I don't see that as critical at the moment, because codecov is reporting anyways.

JBAhire:
https://github.com/hypertrace/actions/pull/9 addresses all changes required on custom actions side.

.github/workflows/pr-build.yml
kotharironak: If we run validate charts as container action, I would suggest we do to publish charts action - `hypertrace/actions/helm-gcs-publish` as well as container actions if there are no challenges for consistency.

.github/workflows/pr-build.yml
aaron-steinfeld: The test reports is meant to report test results (junit) but is being given paths to test coverage (jacoco). Try changing `./**/jacoco/test/*.xml` to `./**/build/test-results/**/*.xml`

aaron-steinfeld: > I don't see that as critical issue at the moment, because codecov is reporting anyways.

It's not critical, but different tools/jobs. If a test fails, codecov will never run, and even if it did, only conveys covered lines of code rather than test results. This step surfaces the results of the tests so a failed test is directly surfaced on the PR rather than buried somewhere in the logs of the test action.

JBAhire: Thanks, this is resolved.

.github/workflows/actions/gradle/action.yml
```diff
@@ -0,0 +1,10 @@
+name: 'Gradle command action'
+inputs:
```
aaron-steinfeld, line 2: Guessing this was here for testing since the workflows are using the common one now

aaron-steinfeld (dismissed):
Please remove the extra action either in this PR or a followup

aaron-steinfeld (approved):

buchi-busireddy (approved):

PR #66, opened 2021-04-16
feat: Log view handler

rish691:
Add LogEventView
hypertrace/hypertrace#224

### DIFF
--- a/query-service/src/main/resources/configs/common/application.conf
+++ b/query-service/src/main/resources/configs/common/application.conf
@@ -285,22 +285,22 @@ service.config = {
           }
         }
       }
-      {
-        name = log-event-handler
-        type = pinot
-        clientConfig = zookeeper
-        clientConfig = ${?PINOT_CONNECTION_TYPE}
-        requestHandlerInfo = {
-          tenantColumnName = tenant_id
-          viewDefinition = {
-            viewName = logEventView
-            bytesFields = ["span_id", "trace_id"]
-            fieldMap = {
-              "LOG_EVENT.timestamp_nanos": "timestamp_nanos",
-              "LOG_EVENT.spanId": "span_id",
-              "LOG_EVENT.traceId": "trace_id",
-              "LOG_EVENT.attributes": "attributes"
-            }
+    }
+    {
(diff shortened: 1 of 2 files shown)


query-service/src/main/resources/configs/common/application.conf
```diff
@@ -285,6 +285,25 @@ service.config = {
...
+        requestHandlerInfo = {
+          tenantColumnName = tenant_id
+          viewDefinition = {
+            viewName = logEventView
+            bytesFields = ["span_id", "trace_id"]
+            fieldMap = {
+              "LOG_EVENT.timestamp_nanos": "timestamp_nanos",
+              "LOG_EVENT.spanId": "span_id",
+              "LOG_EVENT.traceId": "trace_id",
+              "LOG_EVENT.attributes": "attributes"
```
rish691, line 302: I think no extra handling is needed for attributes field which is a json string. 
Currently map is stored in two columns, which is then combined to a Map<String, String> and then converted to a Json string, thus it needs to be specially handled.  
Attribute field is already stored as a json string (with 1 level nesting).

aaron-steinfeld (commented):
int test issue

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -286,6 +286,25 @@ service.config = {
...
+      type = pinot
+      clientConfig = zookeeper
+      clientConfig = ${?PINOT_CONNECTION_TYPE}
+      requestHandlerInfo = {
+        tenantColumnName = tenant_id
+        viewDefinition = {
+          viewName = logEventView
+          bytesFields = ["span_id", "trace_id"]
+          fieldMap = {
+            "LOG_EVENT.timestamp_nanos": "timestamp_nanos",
```
aaron-steinfeld, line 300: Heads up on this line if the attr key changes

kotharironak:
@rish691 Pl. add description and associated embedded log events issue ticket.

aaron-steinfeld (approved):
lgtm pending description

tim-mwangi (approved):

PR #67, opened 2021-04-20
fix: update concat function handling

aaron-steinfeld:
## Description
Adds support for new concat_or_null function, and fixes handling of basic concat to use a custom udf which has semantics matching our concat elsewhere. This should go in after https://github.com/hypertrace/pinot/pull/78

### Testing
Updated UTs

### Checklist:


aaron-steinfeld:
Hmm - will look at this more tomorrow - it's failing directly on main, I think likely due to the fact that the integration tests work exclusively in unpinned versions. Seeing kafka connection errors - @rish691 any idea? Related to the recent merges around log ingestion?

rish691:
> Hmm - will look at this more tomorrow - it's failing directly on main (e.g. https://github.com/hypertrace/query-service/runs/2386946697?check_suite_focus=true ), I think likely due to the fact that the integration tests work exclusively in unpinned versions. Seeing kafka connection errors - @rish691 any idea? Related to the recent merges around log ingestion?

Yeah, this pr should fix the test https://github.com/hypertrace/query-service/pull/69
We could use pinned version also but I guess it will difficult to keep it up to date.

aaron-steinfeld:
> > Hmm - will look at this more tomorrow - it's failing directly on main (e.g. https://github.com/hypertrace/query-service/runs/2386946697?check_suite_focus=true ), I think likely due to the fact that the integration tests work exclusively in unpinned versions. Seeing kafka connection errors - @rish691 any idea? Related to the recent merges around log ingestion?
> 
> Yeah, this pr should fix the test #69
> We could use pinned version also but I guess it will be difficult to keep it up to date.

Thanks for fixing it. In general, my feeling is that tests should always be reproducible, but I also do want them to stay up to date. A bot for updating is one option, but maybe an easier short term solution that could at least assist in debugging is to dump out the versions even if they're not pinned?

avinashkolluru (approved):

tim-mwangi (approved):

rish691:
> but maybe an easier short term solution that could at least assist in debugging is to dump out the versions even if they're not pinned?

I can try that (but the image I pull would only have the main tag 🤔).

aaron-steinfeld:
yeah, true - in that case, maybe dumping the SHA at least would allow us to track back if needed?

PR #73, opened 2021-04-23
Fixed internal error in pinot queries caused due to question mark (?) in query parameter.

singhalprerana:
Pinot PreparedStatement creates invalid query if one of the parameters has '?' in its value.
Sample pinot query : `select * from table where team in (?, ?, ?)`..
Now say parameters are: 'abc', 'pqr with (?)' and 'xyz'..

Now, on executing PreparedStatement:fillStatementWithParameters method on this will return
`select * from table where team in ('abc', 'pqr with ('xyz')', ?)`
-- which is clearly wrong
(what we wanted was `select * from table where team in ('abc', 'pqr with (?)', 'xyz')`)..

The reason is the usage of replaceFirst iteration in the pinot PreparedStatement method..
Hence written a custom method to resolve the query statement rather than relying on Pinot's library method.

Also raised an issue in incubator-pinot github repo: https://github.com/apache/incubator-pinot/issues/6834

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
@@ -345,7 +345,7 @@ public class PinotBasedRequestHandler implements RequestHandler {
       try {
         resultSetGroup =
             pinotQueryExecutionTimer.recordCallable(
-                () -> pinotClient.executeQuery(resolveStatement(pql.getKey(), pql.getValue())));
+                () -> pinotClient.executeQuery(pql.getKey(), pql.getValue()));
       } catch (Exception ex) {
         // Catch this exception to log the Pinot SQL query that caused the issue
         LOG.error("An error occurred while executing: {}", pql.getKey(), ex);
@@ -569,31 +569,5 @@ public class PinotBasedRequestHandler implements RequestHandler {
           noGroupBy && noAggregations,
           "If distinct selections are requested, there should be no groupBys or aggregations.");
     }
-  }
-
-  @VisibleForTesting
-  String resolveStatement(String query, Params params) {
-    String[] queryParts = query.split("\\?");
-
-    String[] parameters = new String[queryParts.length];
-    params.getStringParams().forEach((i, p) -> parameters[i] = getStringParam(p));
-    params.getIntegerParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-    params.getLongParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-    params.getDoubleParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-    params.getFloatParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-    params
-        .getByteStringParams()
-        .forEach((i, p) -> parameters[i] = getStringParam(Hex.encodeHexString(p.toByteArray())));
-
-    StringBuilder sb = new StringBuilder();
-    for (int i = 0; i < queryParts.length; i++) {
-      sb.append(queryParts[i]);
-      sb.append(parameters[i] != null ? parameters[i] : "");
-    }
-    return sb.toString();
-  }
-
-  private String getStringParam(String value) {
-    return "'" + value.replace("'", "''") + "'";
   }
 }
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
@@ -1,6 +1,7 @@
 package org.hypertrace.core.query.service.pinot;
 
-import com.google.common.annotations.VisibleForTesting;
+import static org.hypertrace.core.query.service.ConfigUtils.optionallyGet;
+
 import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Iterables;
@@ -10,7 +11,19 @@ import com.google.protobuf.util.JsonFormat;
 import com.typesafe.config.Config;
 import io.micrometer.core.instrument.Timer;
 import io.reactivex.rxjava3.core.Observable;
-import org.apache.commons.codec.binary.Hex;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
 import org.apache.pinot.client.ResultSet;
 import org.apache.pinot.client.ResultSetGroup;
 import org.hypertrace.core.query.service.ExecutionContext;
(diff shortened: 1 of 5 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -85,6 +85,10 @@ public ResultSetGroup executeQuery(String statement, Params params) {
       return preparedStatement.execute();
     }
 
+    public ResultSetGroup executeQuery(String statement) {
```
rish691, line 88: Instead of adding this method and a new method `PinotBasedRequestHandler#resolveStatement`, why not modify the existing method `buildPreparedStatement` ?

rish691:
nit: jira id can be removed from the description

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
singhalprerana: Wanted to keep that functionality once that gets fixed in pinot repo. But yes we can do that -- pass resolved statement and empty params to prepared statement -- but then the extra processing in PreparedStatement would be redundant.

avinashkolluru: Does it mean the below `executeQueryAsync` method will continue to have the issue.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -568,4 +570,30 @@ private void validateQueryRequest(ExecutionContext executionContext, QueryReques
           "If distinct selections are requested, there should be no groupBys or aggregations.");
     }
   }
+
+  @VisibleForTesting
```
avinashkolluru, line 574: Please add a comment saying this is a temporary fix and also mention the pinot bug here.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
singhalprerana: fixed both the methods.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
singhalprerana: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -102,5 +108,47 @@ private PreparedStatement buildPreparedStatement(String statement, Params params
...
+     * Now, on executing PreparedStatement:fillStatementWithParameters method on this will return
+     * select * from table where team in ('abc', 'pqr with ('xyz')', ?) -- which is clearly wrong
+     * (what we wanted was select * from table where team in ('abc', 'pqr with (?)', 'xyz'))..
+     *
+     * The reason is the usage of replaceFirst iteration in the pinot PreparedStatement method..
+     *
+     * Hence written this custom method to resolve the query statement rather than relying on Pinot's
+     * library method.
+     * This is a temporary fix and can be reverted when the Pinot issue gets resolved
+     * Raised an issue in incubator-pinot github repo: apache/incubator-pinot#6834
```
satish-mittal, line 127: Interesting find!

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -102,5 +108,47 @@ private PreparedStatement buildPreparedStatement(String statement, Params params
...
+     */
+    String resolveStatement(String query, Params params) {
+      String[] queryParts = query.split("\\?");
```
jcchavezs, line 130: Does it make sense to early return is `queryParts` is empty?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -102,5 +108,47 @@ private PreparedStatement buildPreparedStatement(String statement, Params params
...
+     * (what we wanted was select * from table where team in ('abc', 'pqr with (?)', 'xyz'))..
+     *
+     * The reason is the usage of replaceFirst iteration in the pinot PreparedStatement method..
+     *
+     * Hence written this custom method to resolve the query statement rather than relying on Pinot's
+     * library method.
+     * This is a temporary fix and can be reverted when the Pinot issue gets resolved
+     * Raised an issue in incubator-pinot github repo: apache/incubator-pinot#6834
+     */
+    String resolveStatement(String query, Params params) {
```
jcchavezs, line 129: Can this be static? I don't see this using the class state anywhere (maybe I am missing something).

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -102,5 +108,47 @@ private PreparedStatement buildPreparedStatement(String statement, Params params
...
+
+      StringBuilder sb = new StringBuilder();
+      for (int i = 0; i < queryParts.length; i++) {
+        sb.append(queryParts[i]);
+        sb.append(parameters[i] != null ? parameters[i] : "");
+      }
+      return sb.toString();
+    }
+
+    private String getStringParam(String value) {
```
jcchavezs, line 150: I think this can be static.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
jcchavezs, line 129: What happens in this method when the number of `?` don't match the number of params?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -102,5 +108,47 @@ private PreparedStatement buildPreparedStatement(String statement, Params params
           .forEach((i, b) -> preparedStatement.setString(i, Hex.encodeHexString(b.toByteArray())));
       return preparedStatement;
     }
+
+    @VisibleForTesting
+    /*
+     * Pinot PreparedStatement creates invalid query if one of the parameters has '?' in its value.
```
jcchavezs, line 114: nice documentation 💯

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
singhalprerana: if number of params are more, we drop the extras and if it's less, we replace `?` with empty string.. the loop takes care of this..

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
singhalprerana: yep.. fixing..

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
singhalprerana: yep.. fixing..

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
singhalprerana: can do..

avinashkolluru (approved):

rish691 (approved):

PR #78, opened 2021-07-01
feat: update cost of handling request based on request and view definition

GurtejSohi:
## Description
`RequestHandler.canHandle()` returns the query cost. So, if there are more than 1 request handlers that can handle the requested query, then the handler with the min cost is selected to handle that query.

Now, this PR comes up with a way to compute the cost based on request and view definition in the following way (not mentioning the existing checks for simplicity):
```
If (startTime in query < currentTime - viewRetentionTime), then 
  cost is > 0.5   // (exact details mentioned later)
Else
  cost is < 0.5
```
So, there are two classes of request handlers based on this logic - first one with data newer than the requested start time (corresponding to the `if` block), and second one with data older than the requested start time (corresponding to `else` block). 
Request handlers belonging to second class will always be given preference over the request handlers belonging to first class.

Also, within the first class of request handlers, the request handler which has the most overlap(or the max. retention time) with the requested time range is given preference.

And within the second class, the request handler with the finest granularity is given preference.


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -164,10 +173,24 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
...
-    // Finer the time granularity --> Higher the cost.
-    return new QueryCost(0.5);
+//    TODO
+//    if (!this.viewDefinitionSupportsGranularity(viewDefinition, request.getGroupByList())) {
+//      return QueryCost.UNSUPPORTED;
+//    }
+
+    double cost;
+
+    long requestStartTime = getRequestStartTime(request.getFilter());
```
rish691, line 183: I think the logic for computing cost can be abstracted in a method and unit tested?

GurtejSohi: sure, will do that

GurtejSohi: Added unit tests.
Regarding abstracting this in a method - actually, this method itself is for computing the query cost.

rish691 (approved):

PR #79, opened 2021-07-02
Dummy PR for creating a new release

GurtejSohi:
## Description
The tag was mentioned wrongly as `v0.5.12` in the last release (the prefix `v` should not be there). So, creating a dummy PR (adding some docs) so that we can create a new release


GurtejSohi:
Not sure why these tests are failing. The only error that I'm seeing is: 
```
* What went wrong:
No match found
```

Any ideas why this could be failing? Could it be because of the wrong tags of the previous commit in main branch?

kotharironak:
I think, we already have images published -  https://hub.docker.com/layers/hypertrace/query-service/0.5.12/images/sha256-4774aa11cc07a6332b90734e971a31b1e1a72662618c58c1696c87ed31d1ce2e?context=explore

is it possible to manually correct the tag?
cc : @JBAhire

kotharironak (approved):

PR #95, opened 2021-11-10
fix(PinotMapConverter): fill empty string, instead of null

skjindal93:


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotMapConverter.java
```diff
@@ -77,7 +77,7 @@ String merge(String keyData, String valueData) throws IOException {
         map.put(keys.get(idx), values.get(idx));
       } else {
         // to handle unbalanced size
-        map.put(keys.get(idx), null);
+        map.put(keys.get(idx), "");
```
aaron-steinfeld, line 80: Is the able to look up the default null value? Do we have access to that config?

avinashkolluru: Sorry. What config are you referring to?
This change looks fine to me.

aaron-steinfeld: The default value of a column isn't some immutable thing decided by pinot - it's driven by config. For example:
https://github.com/hypertrace/pinot/blob/668e83badb26482af30391888779d754ceff0edf/pinot-servicemanager/schemas/spanEventView-schemaFile.json#L51-L54

We generally write this programmatically based on the avro schema:
https://github.com/hypertrace/view-generator-framework/blob/46edd832fd0765b7b86e593776bc5d34f3780736/view-creator-framework/src/main/java/org/hypertrace/core/viewcreator/pinot/PinotUtils.java#L257

and should be readable if we have access to the schema:
https://github.com/hypertrace/view-generator-framework/blob/46edd832fd0765b7b86e593776bc5d34f3780736/view-creator-framework/src/test/java/org/hypertrace/core/viewcreator/pinot/PinotUtilsTest.java#L60

avinashkolluru: If the pinot schema is automatically derived from the avro schema which has a default of {} for the map<string> fields, I'm really not clear how changing the default value for the case where the number of keys dont match the values from null to `""` is impacted by the pinot schema.

avinashkolluru: Actually dont know the reason for adding this special handling

```
List<String> values;
    if (PINOT_EMPTY_LIST.equals(valueData)) {
      values = new ArrayList<>();
    } else {
      try {
        values = objectMapper.readValue(valueData, listOfString);
      } catch (IOException e) {
        LOG.error(
            "Failed to deserialize map's value to list of string object. Raw Json String {}",
            valueData);
        throw e;
      }
    }
```

aaron-steinfeld: > If the pinot schema is automatically derived from the avro schema which has a default of {} for the map fields

True, this is in the scope of a map, so we can't explicitly set the default for the values of a map. We just need to know that it's a map of string values, since a map<int> would have a different default.

> Actually dont know the reason for adding this special handling

As in why is empty list treated differently? not sure.

avinashkolluru (approved):

aaron-steinfeld (approved):

PR #96, opened 2021-11-15
Added methods for handling custom group by

sarthak77:
Ref issue : https://github.com/hypertrace/hypertrace-ui/issues/1099

### DIFF
--- a/query-service-api/src/main/proto/request.proto
+++ b/query-service-api/src/main/proto/request.proto
@@ -14,7 +14,7 @@ message Expression {
     LiteralConstant literal = 2;
     Function function = 3;
     OrderByExpression orderBy = 4;
-    ObjectIdentifier objectIdentifier = 5;
+    AttributeExpression attributeExpression = 5;
   }
 }
 message QueryRequest {
@@ -74,9 +74,9 @@ message ColumnIdentifier {
   string alias = 2;
 }
 
-message ObjectIdentifier {
-  string columnName = 1;
-  string path_expression = 2;
+message AttributeExpression {
+  string attributeId = 1;
+  string subpath = 2;
   string alias = 3;
 }
 
--- a/query-service-api/src/main/proto/request.proto
+++ b/query-service-api/src/main/proto/request.proto
@@ -76,8 +76,8 @@ message ColumnIdentifier {
 
 message AttributeExpression {
   string attributeId = 1;
-  string subpath = 2;
-  string alias = 3;
+  optional string subpath = 2;
+  optional string alias = 3;
 }
 
 message OrderByExpression {
--- a/query-service-api/src/main/proto/request.proto
+++ b/query-service-api/src/main/proto/request.proto
@@ -14,6 +14,7 @@ message Expression {
     LiteralConstant literal = 2;
     Function function = 3;
     OrderByExpression orderBy = 4;
+    ObjectIdentifier objectIdentifier = 5;
   }
 }
 message QueryRequest {
@@ -71,6 +72,12 @@ message ColumnIdentifier {
 
   string columnName = 1;
   string alias = 2;
+}
+
+message ObjectIdentifier {
+  string columnName = 1;
+  string path_expression = 2;
+  string alias = 3;
 }
 
 message OrderByExpression {
--- a/query-service-api/src/main/proto/request.proto
+++ b/query-service-api/src/main/proto/request.proto
@@ -14,7 +14,7 @@ message Expression {
     LiteralConstant literal = 2;
     Function function = 3;
     OrderByExpression orderBy = 4;
-    AttributeExpression attributeExpression = 5;
+    AttributeExpression attribute_expression = 5;
   }
 }
 message QueryRequest {
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -1,6 +1,6 @@
 package org.hypertrace.core.query.service.pinot;
 
-import static org.hypertrace.core.query.service.api.Expression.ValueCase.ATTRIBUTEEXPRESSION;
+import static org.hypertrace.core.query.service.api.Expression.ValueCase.ATTRIBUTE_EXPRESSION;
 import static org.hypertrace.core.query.service.api.Expression.ValueCase.COLUMNIDENTIFIER;
 import static org.hypertrace.core.query.service.api.Expression.ValueCase.LITERAL;
 
@@ -8,7 +8,6 @@ import com.google.common.base.Joiner;
 import com.google.protobuf.ByteString;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Map.Entry;
 import org.hypertrace.core.query.service.ExecutionContext;
 import org.hypertrace.core.query.service.api.Expression;
@@ -162,7 +161,7 @@ class QueryRequestToPinotSQLConverter {
         case CONTAINS_KEYVALUE:
           kvp = convertExpressionToMapLiterals(filter.getRhs());
           String logicalColumnName = getLogicalColumnName(filter.getLhs());
-          if (filter.getLhs().getValueCase() == ATTRIBUTEEXPRESSION) {
+          if (filter.getLhs().getValueCase() == ATTRIBUTE_EXPRESSION) {
             String pathExpression = filter.getLhs().getAttributeExpression().getSubpath();
             kvp[0] =
                 LiteralConstant.newBuilder()
@@ -277,11 +276,9 @@ class QueryRequestToPinotSQLConverter {
       Expression expression, Builder paramsBuilder, ExecutionContext executionContext) {
     switch (expression.getValueCase()) {
       case COLUMNIDENTIFIER:
-      case ATTRIBUTEEXPRESSION:
+      case ATTRIBUTE_EXPRESSION:
         // this takes care of the Map Type where it's split into 2 columns
-        List<String> columnNames =
-            viewDefinition.getPhysicalColumnNames(getLogicalColumnName(expression));
-        return joiner.join(columnNames);
+        return joiner.join(viewDefinition.getPhysicalColumnNames(getLogicalColumnName(expression)));
       case LITERAL:
         return convertLiteralToString(expression.getLiteral(), paramsBuilder);
       case FUNCTION:
@@ -303,7 +300,7 @@ class QueryRequestToPinotSQLConverter {
     switch (expression.getValueCase()) {
       case COLUMNIDENTIFIER:
         return expression.getColumnIdentifier().getColumnName();
-      case ATTRIBUTEEXPRESSION:
+      case ATTRIBUTE_EXPRESSION:
         return expression.getAttributeExpression().getAttributeId();
       default:
         throw new IllegalArgumentException(
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -160,47 +160,31 @@ class QueryRequestToPinotSQLConverter {
           break;
         case CONTAINS_KEYVALUE:
           kvp = convertExpressionToMapLiterals(filter.getRhs());
-          String logicalColumnName = getLogicalColumnName(filter.getLhs());
+          String keyCol = convertStringToMapKeyColumn(getLogicalColumnName(filter.getLhs()));
+          String valCol = convertStringToMapValueColumn(getLogicalColumnName(filter.getLhs()));
+          builder.append(getStringForMapAttribute(kvp, keyCol, valCol, paramsBuilder));
+          break;
+        default:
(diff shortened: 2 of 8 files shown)


query-service-api/src/main/proto/request.proto
```diff
@@ -14,6 +14,7 @@ message Expression {
     LiteralConstant literal = 2;
     Function function = 3;
     OrderByExpression orderBy = 4;
+    ObjectIdentifier objectIdentifier = 5;
```
aaron-steinfeld, line 17: This is against the lint rules, but I guess consistency is better.

query-service-api/src/main/proto/request.proto
```diff
@@ -73,6 +74,12 @@ message ColumnIdentifier {
   string alias = 2;
 }
 
+message ObjectIdentifier {
```
aaron-steinfeld, line 77: new message should use consistent snake_casing.

Also does this replace column identifier? Do we deprecate that?

aaron-steinfeld (commented):
Just commenting on the API for now, but let's validate the approach before proceeding as discussed in https://github.com/hypertrace/hypertrace-ui/issues/1099#issuecomment-968991765

query-service-api/src/main/proto/request.proto
```diff
@@ -73,6 +74,12 @@ message ColumnIdentifier {
   string alias = 2;
 }
 
+message AttributeExpression {
+  string attributeId = 1;
```
aaron-steinfeld, line 78:
```suggestion
  string attribute_id = 1;
```

(snake_case is our platform standard; normally I'd say be consistent, but given the difficulty of changing these later I think we should use correct casing on all new things)

query-service-api/src/main/proto/request.proto
```diff
@@ -73,6 +74,12 @@ message ColumnIdentifier {
...
+  string subpath = 2;
```
aaron-steinfeld, line 79: optional
```suggestion
  optional string subpath = 2;
```

query-service-api/src/main/proto/request.proto
```diff
@@ -73,6 +74,12 @@ message ColumnIdentifier {
...
+  string alias = 3;
```
aaron-steinfeld, line 80:
```suggestion
  optional string alias = 3;
```

query-service-api/src/main/proto/request.proto
```diff
@@ -14,6 +14,7 @@ message Expression {
     LiteralConstant literal = 2;
     Function function = 3;
     OrderByExpression orderBy = 4;
+    AttributeExpression attributeExpression = 5;
```
aaron-steinfeld, line 17:
```suggestion
    AttributeExpression attribute_expression = 5;
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -263,9 +277,13 @@ private String convertExpression2String(
...
       case COLUMNIDENTIFIER:
-        String logicalColumnName = expression.getColumnIdentifier().getColumnName();
         // this takes care of the Map Type where it's split into 2 columns
-        List<String> columnNames = viewDefinition.getPhysicalColumnNames(logicalColumnName);
+        List<String> columnNames =
+            viewDefinition.getPhysicalColumnNames(getLogicalColumnName(expression));
+        return joiner.join(columnNames);
+      case ATTRIBUTEEXPRESSION:
+        // this takes care of the Map Type where it's split into 2 columns
+        columnNames = viewDefinition.getPhysicalColumnNames(getLogicalColumnName(expression));
```
aaron-steinfeld, line 286: Bad practice to share variables between cases. I'm actually not even sure what java does in this case since control flow never hits the declaration on line 281 - probably doesn't care, but I'd be curious. Either way, best avoided. Just inline columnNames inside the join fn.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -284,25 +302,31 @@ private String convertExpression2String(
...
+        return expression.getColumnIdentifier().getColumnName();
+      case ATTRIBUTEEXPRESSION:
+        return expression.getAttributeExpression().getAttributeId();
+      default:
+        throw new IllegalArgumentException(
+            "operator CONTAINS_KEY/KEYVALUE supports multi value column only");
+    }
+  }
+
+  private String convertExpressionToMapKeyColumn(String logicalColumnName) {
```
aaron-steinfeld, line 317: No longer taking an expression, please rename

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -320,10 +346,17 @@ private String convertExpressionToMapValueColumn(Expression expression) {
                       Value.newBuilder().setString(value.getValue().getStringArray(i)).build())
                   .build();
         }
+      } else if (value.getValue().getValueType() == ValueType.STRING) {
+        literals[1] =
```
aaron-steinfeld, line 350: So literals `[0]` is null? We should return something better (but see my comment about operator support, may make this moot)

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -154,14 +155,27 @@ private String convertFilter2String(
           break;
         case CONTAINS_KEY:
           LiteralConstant[] kvp = convertExpressionToMapLiterals(filter.getRhs());
-          builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
+          builder.append(convertExpressionToMapKeyColumn(getLogicalColumnName(filter.getLhs())));
```
aaron-steinfeld, line 158: I think we're doing the opposite of what we want here. CONTAINS_KEY and CONTAINS_KEYVALUE operators are defunct - attribute expressions replace them, and it doesn't make sense to use the two together. In both of these operators, the path is specified as an operator argument (RHS) - attribute expression moves path to the LHS in order to allow it to be used with any operator. These two special cased ones would go away, since it would never make sense to have something like `AttributeExpression("myCol", "my.path") CONTAINS_KEY_VALUE ("my.path", "my.value")` - we'd simply say `AttributeExpression("myCol", "my.path") EQUALS "my.value"`

query-service-api/src/main/proto/request.proto
sarthak77: Aren't we using camel case everywhere ?

query-service-api/src/main/proto/request.proto
sarthak77: done

query-service-api/src/main/proto/request.proto
sarthak77: Done.
Isn't it optional by default ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: Java gives error if we declare the same variable in different `CASE` statements. 
Quoted from online sources : 
"
That's because a case doesn't create a scope. So, both the variables in the 2 cases are in the same scope. If you want to do this, you can add braces for each case, which will create a new scope for each case.
"
Combined both the cases as of now.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: done

query-service-api/src/main/proto/request.proto
aaron-steinfeld: this file is breaking our coding standards enforced in every other repo (or most of them). snake_case is our standard for proto messages, we follow the default lint style from https://docs.buf.build/lint/rules#default .

query-service-api/src/main/proto/request.proto
aaron-steinfeld: Any field can be omitted, but optional allows for field presence checking. Any primitive field that we expect to be omitted in a reasonable operation, and want the consumer to _know_ we omitted it should be marked optional. Explained in detail at https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md

query-service-api/src/main/proto/request.proto
sarthak77: resolved

query-service-api/src/main/proto/request.proto
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
aaron-steinfeld: Yes, you can't redeclare the same variable between case statements - but there's no need for a variable here, it's only used on the following line and doesn't add clarity.

query-service-api/src/main/proto/request.proto
sarthak77: resolved

query-service-api/src/main/proto/request.proto
sarthak77: resolved

query-service-api/src/main/proto/request.proto
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: resolved

sarthak77:
Merged proto changes and snyk file changes as a part of another PR (https://github.com/hypertrace/query-service/pull/98) .

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: Returning empty string instead

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -154,40 +154,37 @@ private String convertFilter2String(
...
           break;
         default:
-          rhs = handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
-          builder.append(
-              convertExpression2String(filter.getLhs(), paramsBuilder, executionContext));
-          builder.append(" ");
-          builder.append(operator);
-          builder.append(" ");
-          builder.append(convertExpression2String(rhs, paramsBuilder, executionContext));
+          if (filter.getLhs().getValueCase() == ATTRIBUTE_EXPRESSION) {
```
aaron-steinfeld, line 168: If we need to handle this across each path, we'll run into bugs (does this support the like case above, for example?). Can we have one spot where we just do a general conversion of expressions that all paths (filter, select, group, order) go through?

sarthak77: This PR was just for `CONTAINS_KEYVALUE`. Other things would be added in the next PR (Ref : https://github.com/hypertrace/query-service/pull/97)

aaron-steinfeld: Discussed this a bit offline, but this is what I meant in https://github.com/hypertrace/query-service/pull/96#discussion_r750649816 . If we never add attribute expression support for CONTAINS_KEY and CONTAINS_KEYVALUE, that's fine - those are meant to be replaced by attribute expressions anyway. What we need to do in this PR (or in multiple) is everywhere we resolve an `Expression` into a logical column name add support for that `Expression` to either be of type `ColumnIdentifier` (e.g. `TRACE.id -> id`) or of type `AttributeExpression`. AttributeExpression could be a simple translation like `TRACE.id -> id`, or a complex translation like `SPAN.tags w/ subpath span.kind -> mapValue(tags__KEYS, 'span.kind' ,tags__VALUES)`

This is what gives us full support for referencing custom keys across all operations

aaron-steinfeld: We also don't want to do this ad-hoc in each spot, so this should happen inside `convertExpression2String`, or something like it (who names with numbers!?)

sarthak77:
Closing this PR. All necessary changes have been added in a separate PR.

PR #102, opened 2021-11-19
fix : sorting rate metrics

sarthak77:
On the service (backend) screen page, there is a rate column - `error/s`. After we moved the avgrate functionality to query-service, the sorting on this column is broken.

The reason it was broken, that avgrate function takes two arguments (column name, size). The second size argument is missing in `order by` expression. Priori, it was working fine as `avgrate` was translated to `sum` at the gateway service layer. And `sum` function needs only single column name argument.

As these rate columns are per second rate, for now, defaulting to `PT1S` as size for avgrate function to fix the regression.

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
@@ -28,6 +28,7 @@ public class PinotFunctionConverter {
   private static final String DEFAULT_PERCENTILE_AGGREGATION_FUNCTION = "PERCENTILETDIGEST";
 
   private static final String PINOT_CONCAT_FUNCTION = "CONCATSKIPNULL";
+  private static final String DEFAULT_AVG_RATE_SIZE = "PT1S";
 
   private final String percentileAggFunction;
 
@@ -80,8 +81,11 @@ public class PinotFunctionConverter {
       java.util.function.Function<Expression, String> argumentConverter,
       ExecutionContext executionContext) {
 
-    String columnName = argumentConverter.apply(getColumnNameForAvgRate(function).get());
-    String rateIntervalInIso = getRateIntervalForAvgRate(function);
+    String columnName = argumentConverter.apply(function.getArgumentsList().get(0));
+    String rateIntervalInIso =
+        function.getArgumentsList().size() == 2
+            ? function.getArgumentsList().get(1).getLiteral().getValue().getString()
+            : DEFAULT_AVG_RATE_SIZE;
     long rateIntervalInSeconds = isoDurationToSeconds(rateIntervalInIso);
     long aggregateIntervalInSeconds =
         (executionContext
@@ -93,18 +97,6 @@ public class PinotFunctionConverter {
     return String.format(
         "SUM(DIV(%s, %s))",
         columnName, (double) aggregateIntervalInSeconds / rateIntervalInSeconds);
-  }
-
-  private Optional<Expression> getColumnNameForAvgRate(Function function) {
-    return function.getArgumentsList().stream().filter(e -> e.hasColumnIdentifier()).findFirst();
-  }
-
-  private String getRateIntervalForAvgRate(Function function) {
-    return function.getArgumentsList().stream()
-        .filter(e -> e.hasLiteral())
-        .map(e -> e.getLiteral().getValue().getString())
-        .findFirst()
-        .orElse("PT1S");
   }
 
   private String convertCount() {
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
@@ -81,7 +81,7 @@ public class PinotFunctionConverter {
       ExecutionContext executionContext) {
 
     String columnName = argumentConverter.apply(getColumnNameForAvgRate(function).get());
-    String rateIntervalInIso = getAvgRateLiteral(function);
+    String rateIntervalInIso = getRateIntervalForAvgRate(function);
     long rateIntervalInSeconds = isoDurationToSeconds(rateIntervalInIso);
     long aggregateIntervalInSeconds =
         (executionContext
@@ -99,7 +99,7 @@ public class PinotFunctionConverter {
     return function.getArgumentsList().stream().filter(e -> e.hasColumnIdentifier()).findFirst();
   }
 
-  private String getAvgRateLiteral(Function function) {
+  private String getRateIntervalForAvgRate(Function function) {
     return function.getArgumentsList().stream()
         .filter(e -> e.hasLiteral())
         .map(e -> e.getLiteral().getValue().getString())
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
@@ -80,9 +80,8 @@ public class PinotFunctionConverter {
       java.util.function.Function<Expression, String> argumentConverter,
       ExecutionContext executionContext) {
 
-    String columnName = argumentConverter.apply(function.getArgumentsList().get(0));
-    String rateIntervalInIso =
-        function.getArgumentsList().get(1).getLiteral().getValue().getString();
+    String columnName = argumentConverter.apply(getColumnNameForAvgRate(function).get());
+    String rateIntervalInIso = getAvgRateLiteral(function);
     long rateIntervalInSeconds = isoDurationToSeconds(rateIntervalInIso);
     long aggregateIntervalInSeconds =
         (executionContext
@@ -94,6 +93,18 @@ public class PinotFunctionConverter {
     return String.format(
         "SUM(DIV(%s, %s))",
         columnName, (double) aggregateIntervalInSeconds / rateIntervalInSeconds);
+  }
+
+  private Optional<Expression> getColumnNameForAvgRate(Function function) {
+    return function.getArgumentsList().stream().filter(e -> e.hasColumnIdentifier()).findFirst();
+  }
+
+  private String getAvgRateLiteral(Function function) {
+    return function.getArgumentsList().stream()
+        .filter(e -> e.hasLiteral())
+        .map(e -> e.getLiteral().getValue().getString())
+        .findFirst()
+        .orElse("PT1S");
   }
 
   private String convertCount() {
--- a/query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
+++ b/query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
@@ -8,12 +8,16 @@ import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createS
 import static org.hypertrace.core.query.service.QueryRequestUtil.createNullNumberLiteralExpression;
 import static org.hypertrace.core.query.service.QueryRequestUtil.createNullStringLiteralExpression;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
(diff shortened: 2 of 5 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -96,6 +95,18 @@ private String functionToStringForAvgRate(
         columnName, (double) aggregateIntervalInSeconds / rateIntervalInSeconds);
   }
 
+  private Optional<Expression> getColumnNameForAvgRate(Function function) {
+    return function.getArgumentsList().stream().filter(e -> e.hasColumnIdentifier()).findFirst();
+  }
+
+  private String getAvgRateLiteral(Function function) {
```
kotharironak, line 102: `getAvgRateLiteral` -> `getLiteralForAvgRate`?

query-service-impl/src/test/resources/service_request_handler.conf
```diff
@@ -0,0 +1,28 @@
+{
```
kotharironak, line 1: Can you add tests ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
sarthak77: `getRateIntervalForAvgRate` ?

query-service-impl/src/test/resources/service_request_handler.conf
sarthak77: added

kotharironak (commented):
lgtm

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -96,6 +95,18 @@ private String functionToStringForAvgRate(
         columnName, (double) aggregateIntervalInSeconds / rateIntervalInSeconds);
   }
 
+  private Optional<Expression> getColumnNameForAvgRate(Function function) {
```
avinashkolluru, line 98: Why are we returning an Optional here if the caller is forcefully doing get()?
Optional here makes sense if the caller needs to deal with the value being present or not.

aaron-steinfeld: Agreed. Also, this isn't returning a column name, it's returning an expression. If I sent my args out of order, this would still work which is not what we want. I would just change the old line 84 to default if empty, and leave the rest as is.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -80,9 +80,8 @@ private String functionToStringForAvgRate(
       java.util.function.Function<Expression, String> argumentConverter,
       ExecutionContext executionContext) {
 
-    String columnName = argumentConverter.apply(function.getArgumentsList().get(0));
-    String rateIntervalInIso =
-        function.getArgumentsList().get(1).getLiteral().getValue().getString();
```
aaron-steinfeld, line 85: That is, change this to something equivalent to 
```java
  function.getArgumentsList().size() == 2 
    ? function.getArgumentsList().get(1).getLiteral().getValue().getString() : DEFAULT_AVG_RATE_SIZE;
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -96,6 +95,18 @@ private String functionToStringForAvgRate(
         columnName, (double) aggregateIntervalInSeconds / rateIntervalInSeconds);
   }
 
+  private Optional<Expression> getColumnNameForAvgRate(Function function) {
+    return function.getArgumentsList().stream().filter(e -> e.hasColumnIdentifier()).findFirst();
+  }
+
+  private String getRateIntervalForAvgRate(Function function) {
```
avinashkolluru, line 102: We have integration-tests in query-service. We should definitely add one for this case.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
sarthak77: resolved

query-service-impl/src/test/resources/service_request_handler.conf
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
sarthak77: made the change

sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
sarthak77: We already have two integration tests for testing the average rate changes. Should I add for this specific case also where average rate is in `order-by` expression ?

avinashkolluru: Yes. We should definitely add one.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -1120,14 +1242,19 @@ private QueryRequest buildSimpleQueryWithFilter(Filter filter) {
     return builder.build();
   }
 
-  private void assertPQLQuery(QueryRequest queryRequest, String expectedQuery) {
+  private void assertPQLQuery(
+      QueryRequest queryRequest, String expectedQuery, ViewDefinition viewDefinition) {
```
aaron-steinfeld, line 1246: if we're going to require passing this in (which is fine), we shouldn't make it a global static.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -1120,14 +1242,19 @@ private QueryRequest buildSimpleQueryWithFilter(Filter filter) {
...
+
+    when(mockingExecutionContext.getTenantId()).thenReturn("__default");
```
aaron-steinfeld, line 1248: An assertion shouldn't do mocking of its own. I would guess these are not used by most test cases, which means mockito would be erroring out unless we're running in lenient mode (I'm guessing we are). We don't need to switch lenient off as part of this change, but we should be strict for any new tests we write.

https://javadoc.io/static/org.mockito/mockito-core/3.2.4/org/mockito/quality/Strictness.html#STRICT_STUBS

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -908,7 +958,24 @@ public void testQueryWithNulls() {
             + viewDefinition.getTenantIdColumn()
             + " = '"
             + TENANT_ID
-            + "' limit 15");
+            + "' limit 15",
+        viewDefinition);
+  }
+
+  @Test
+  public void testQueryWithAverageRateInOrderBy() {
```
sarthak77, line 966: @avinashkolluru the integration test would be exactly the same as this. Should I keep this test also or remove this one?

kotharironak: I think we should keep this as well. It is checking the expected PQL query.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
sarthak77: resolved

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
sarthak77: This was because we need service view definition for testing avgrate changes. Rest all use this same view definition.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
sarthak77: `getTenantId()` is used by most of the tests.

sarthak77: moved others to the respective test.

aaron-steinfeld: we shouldn't share the state though - that's the same problem you ran into with view definition. Have each test pass in their own execution context, same as view def, then they can mock that locally, according to their test parameters. For backwards compat, you could overload the assert method with the existing behavior if you want - that way no existing tests hve to be modified. In the current for, we have a mix of both things we don't want - shared state (including newly introduced shared state) and modifying the existing tests.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
sarthak77: resolved

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
sarthak77: resolved

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -46,27 +48,21 @@
...
-  }
-
   @BeforeEach
   public void setup() {
-    connection = Mockito.mock(Connection.class);
+    connection = mock(Connection.class);
     Mockito.when(connection.prepareStatement(any(Request.class))).thenCallRealMethod();
+    mockingExecutionContext = mock(ExecutionContext.class);
+    when(mockingExecutionContext.getTenantId()).thenReturn("__default");
+    when(this.mockingExecutionContext.getTimeSeriesPeriod()).thenReturn(Optional.empty());
```
aaron-steinfeld, line 63: this mock is not relevant to every test and would fail in strict mode. please move it such that we only mock things that are relevant to the test.

For future tests (beyond the scope of this PR), make sure to always run in strict mode. The easiest way to do that is probably use the junit extension -
```java
@ExtendWith(MockitoExtension.class)
```

aaron-steinfeld (approved):
approving to unblock the bug fix, but we should get the newly added tests cleaned up here or in a followup.

sarthak77:
Raised a follow-up PR for fixing the unit tests : https://github.com/hypertrace/query-service/pull/103

PR #103, opened 2021-11-22
Fix : Updated unit tests

sarthak77:
Updated unit tests to resolve mockito issues

### DIFF
--- a/query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
+++ b/query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
@@ -5,11 +5,15 @@ import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createA
 import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createColumnExpression;
 import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createCountByColumnSelection;
 import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createEqualsFilter;
+import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createFunctionExpression;
+import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createInFilter;
+import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createLongLiteralValueExpression;
 import static org.hypertrace.core.query.service.QueryRequestBuilderUtils.createNotEqualsFilter;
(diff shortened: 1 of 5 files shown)


query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -1013,6 +1004,19 @@ private Filter createTimeFilter(String columnName, Operator op, long value) {
     return Filter.newBuilder().setLhs(lhs).setOperator(op).setRhs(rhs).build();
   }
 
+  private Filter createLongTimeFilter(String columnName, Operator op, long value) {
```
aaron-steinfeld, line 1007: this is just a long filter, there's nothing time specific about it. That said, as with most of these private methods, I'd bet the test utils class already has this functionality ready to go, and this. can be removed.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
```diff
@@ -1268,9 +1272,11 @@ private void assertPQLQuery(
       QueryRequest queryRequest, String expectedQuery, ViewDefinition viewDefinition) {
     QueryRequestToPinotSQLConverter converter =
         new QueryRequestToPinotSQLConverter(viewDefinition, new PinotFunctionConverter());
+    ExecutionContext executionContext = new ExecutionContext("__default", queryRequest);
+    executionContext.setTimeFilterColumn("SERVICE.startTime");
```
aaron-steinfeld, line 1276: removing the mock and replacing it with the real execution context increases the scope of the test - it's no longer a unit test, since it's relying on logic inside the execution context. Is that what you intended?

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
sarthak77: Refactoring some of the utility functions used here.

sarthak77: resolved

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverterTest.java
sarthak77: resolved

aaron-steinfeld (approved):

Issue #104, opened 2021-11-25
Add and test support for handling simple `attributeExpression`


sarthak77:
In response to [this](https://github.com/hypertrace/hypertrace-ui/issues/1099#issuecomment-970211104) discussion, adding support for simple attribute. Eventually, this will replace the `columnIdentifier` but currently adding support so that requests containing any of the `columnIentifier` and `attributeExpression` will work.

e.g In a correct request, a subpath will not be present and attribute is mapped to a **simple** long column, QS can serve it.

```
AttributeExpression {
attributeId: API_TRACE.startTime";
}
```
Dividing it into following sub tasks :

- [x] Add support for simple attributes in `QueryRequestToPinotSQLConverter.java`
- [x] Add relevant tests in `QueryRequestToPinotSQLConverterTest.java`

PR #106, opened 2021-11-25
Added support for handling simple attribute expression

sarthak77:
Ref issue : https://github.com/hypertrace/query-service/issues/104

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -1,5 +1,6 @@
 package org.hypertrace.core.query.service.pinot;
 
+import static org.hypertrace.core.query.service.api.Expression.ValueCase.ATTRIBUTE_EXPRESSION;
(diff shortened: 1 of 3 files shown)


rish691 (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -285,9 +296,9 @@ private String convertExpression2String(
   }
 
   private String convertExpressionToMapKeyColumn(Expression expression) {
-    if (expression.getValueCase() == COLUMNIDENTIFIER) {
-      String logicalColumnName = expression.getColumnIdentifier().getColumnName();
-      String col = viewDefinition.getKeyColumnNameForMap(logicalColumnName);
+    if ((expression.getValueCase() == COLUMNIDENTIFIER)
+        || (expression.getValueCase() == ATTRIBUTE_EXPRESSION && !isComplexAttribute(expression))) {
```
rish691, line 300: nit: `(expression.getValueCase() == ATTRIBUTE_EXPRESSION && !isComplexAttribute(expression))` could be extracted into a method

sarthak77: resolved

PR #110, opened 2021-11-29
feat: promql rest client and response parser for instant and range query

kotharironak:
This PR is for handling the ticket - https://github.com/hypertrace/hypertrace/issues/322

Adds:
- Adds PromQLQuery object both for Instant and Range query of PromQL API (https://prometheus.io/docs/prometheus/latest/querying/api/)
- Adds the PromQL API response parser for Instant and Range requests.
- Adds required unit tests
- Adds support for executing multiple PromQL queries in async mode and waiting for results. This is required when we have multiple metrics requested in a single QS request. As an example, in the below instant metric query request ask for two metrics - `numCall` and `erroCount`. As in Prometheus, they are two different metrics, we need to fetch them in two different requests. More query details - https://github.com/hypertrace/hypertrace/issues/295
```
filter {
  childFilter {
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: GE
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631177908474
          }
        }
      }
    }
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: LT
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631181508474
          }
        }
      }
    }
  }
}
selection {
  columnIdentifier {
    columnName: "BACKEND.id"
  }
}
selection {
  function {
    functionName: "SUM"
    arguments {
      columnIdentifier {
        columnName: "BACKEND.errorCount"
      }
    }
    alias: "SUM_BACKEND.errorCount_[]"
  }
}
selection {
  function {
    functionName: "SUM"
    arguments {
      columnIdentifier {
        columnName: "BACKEND.numCalls"
      }
    }
    alias: "SUM_BACKEND.numCalls_[]"
  }
}
groupBy {
  columnIdentifier {
    columnName: "BACKEND.id"
  }
}
limit: 10000
```
As for the above query, there are two different metrics requested, it will be translated to two queries as an example:
```
sum by (backend_id) 
(sum_over_time(num_calls{tenant_id="__default"}[3600000ms])) 
@ 1631181508474

sum by (backend_id) 
(sum_over_time(error_count{tenant_id="__default"'}[3600000ms])) 
@ 1631181508474
```


query-service-impl/src/main/java/org/hypertrace/core/query/service/promql/PromQLQuery.java
```diff
@@ -0,0 +1,98 @@
+package org.hypertrace.core.query.service.promql;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PromQLQuery {
```
kotharironak, line 6: Need to discuss on a final object with @rish691 here

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
```diff
@@ -0,0 +1,39 @@
...
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class PromQLQuery {
+
+  @NonNull @Singular private List<String> queries;
+  private boolean isInstantRequest;
+  private Instant startTime;
+  private Instant endTime;
+  private Duration step;
+
+  static class Builder {
```
kotharironak, line 24: @rish691 Let me know if you think, we want to divide this into two Query objects - one for Instant and another for RangeQuery.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
```diff
@@ -0,0 +1,114 @@
...
+            .map(
+                request -> {
+                  Call call = okHttpClient.newCall(request);
+                  OkHttpResponseCallback callback = new OkHttpResponseCallback();
+                  call.enqueue(callback);
+                  return callback.future;
+                })
+            .toArray(CompletableFuture[]::new);
+
+    CompletableFuture.allOf(completableFutures).join();
```
kotharironak, line 46: Doing this way, as want to throw an exception if any of the requests fails, would `thanApply` do the same?

Any suggestion here of doing a better way?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusMetricQueryResponseParser.java
```diff
@@ -0,0 +1,162 @@
...
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+
+public class PrometheusMetricQueryResponseParser {
+
+  public static PrometheusMetricQueryResponse parse(String jsonString) {
```
kotharironak, line 12: I did this way as I had to parse value array which is timestamp and string in response (https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries) So, thought of organizing data a bit and so parsed fully. 

Pl. let me know if you guys are thinking differently.

kotharironak:
Here, it says that build `test` is passing - https://github.com/hypertrace/query-service/runs/4393503823?check_suite_focus=true, but that is why it says `waiting for status`. Will check it out.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusMetricQueryResponseParser.java
kotharironak: Moved to Jackson-based parser using object mapper for simplicity (mapped to the same set of classes required from API).

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
```diff
@@ -0,0 +1,97 @@
...
+    @JsonProperty("result")
+    private List<PromQLMetricResult> result;
+  }
+
+  @Getter
+  @Builder
+  @NoArgsConstructor
+  @AllArgsConstructor
+  @EqualsAndHashCode
+  public static class PromQLMetricResult {
```
kotharironak, line 52: Pl. let me know, can move them to individual classes. Added here for ease of reading with https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.List;
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Singular;
+
+@Getter
```
aaron-steinfeld, line 24: Sounds like we want a `Value`

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
```diff
@@ -0,0 +1,97 @@
...
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class PromQLMetricResponse {
```
aaron-steinfeld, line 28: Need to be public? guessing we're encapsulating the prometheus support in this package?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
```diff
@@ -0,0 +1,97 @@
...
+    @JsonDeserialize(using = PromQLMetricValuesDeserializer.class)
+    @Singular
+    private List<PromQLMetricValue> values;
+  }
+
+  @AllArgsConstructor
+  @Getter
+  @EqualsAndHashCode
+  public static class PromQLMetricValue {
+    private long timeStamp;
```
aaron-steinfeld, line 67: Can we make this an instant (preferable) or if there's some reason we cant, at least rename it to include units (also timestamp is one word, lowercase s please)

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
aaron-steinfeld: A boolean that determines behavior sounds like it should be two objects. Guessing only one of the times is used by an instant query, but it's not clear which one (edit: looking below looks like its end time, which requires reading the impl)?

Although I guess the broader question I have - what do instant queries mean in the context of our current query language? I probably missed this conversation, but I'm curious if this is actually required for parity.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
```diff
@@ -0,0 +1,122 @@
...
+
+public class PrometheusRestClient {
+  private static final String INSTANT_QUERY = "api/v1/query";
+  private static final String RANGE_QUERY = "api/v1/query_range";
+
+  private String host;
+  private int port;
+  private OkHttpClient okHttpClient;
+
+  public PrometheusRestClient(String host, int port, OkHttpClient okHttpClient) {
```
aaron-steinfeld, line 24: the transport should be internal to the client

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
```diff
@@ -0,0 +1,122 @@
...
+    this.okHttpClient = okHttpClient;
+    this.host = host;
+    this.port = port;
+  }
+
+  public List<ImmutablePair<Request, PromQLMetricResponse>> execute(PromQLQuery query) {
```
aaron-steinfeld, line 30: Why return a list of pairs instead of a map?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
kotharironak: >Although I guess the broader question I have - what do instant queries mean in the context of our current query language?

Aggregation metric query without `dataTimeConvert` group by clause (non-time series query), but asking for metric value for a given point in time. e.g error_count values on last 15 mins on the metric page (above time series, we are showing Total). Query Service example in description note.

e.g PQL query:
```
Select service_id, sum(numCalls) 
FROM spanEventView 
WHERE tenant_id = '__default' 
AND ( service_id != 'null' 
AND ( start_time_millis >= 1631175154903 AND start_time_millis < 1631178754903 ) 
AND ( service_id = '27857c9c-f6fa-30fe-ac1d-48fd07374775' ) ) 
GROUP BY service_id 
limit 1
```

kotharironak: Discussed with @rish691, breaking PromQLQuery object into two PromQLInstantQuery and PromQLRangeQuery.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
```diff
@@ -0,0 +1,97 @@
...
+      if (node.isArray() && node.get(0).isArray()) {
+        node.elements().forEachRemaining(e -> metricValues.add(parseValue((ArrayNode) e)));
+      } else {
+        metricValues.add(parseValue((ArrayNode) node));
+      }
+      return metricValues;
+    }
+
+    private PromQLMetricValue parseValue(ArrayNode arrayNode) {
+      long timestamp = (long) arrayNode.get(0).asDouble() * 1000L;
```
sarthak77, line 88: nit : rename timestamp to something which conveys the unit ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
```diff
@@ -0,0 +1,39 @@
...
+@Getter
+@Builder(builderClassName = "Builder", buildMethodName = "build")
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class PromQLQuery {
+
+  @NonNull @Singular private List<String> queries;
+  private boolean isInstantRequest;
+  private Instant startTime;
+  private Instant endTime;
+  private Duration step;
```
sarthak77, line 22: what is step ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
```diff
@@ -0,0 +1,39 @@
...
+
+  static class Builder {
+    PromQLQuery build() {
+      if (queries.size() == 0) {
+        throw new UnsupportedOperationException(
+            "query string requires for Instant or Range query request");
```
sarthak77, line 28: nit : requires -> required

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
kotharironak: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
kotharironak: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
kotharironak: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
kotharironak: I was thinking, it will be easy during converting to QS response. But, we can go with Instant. Made the changes.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
kotharironak: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
kotharironak: Change it to `Instant` type.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
kotharironak: Here also, I was thinking, while building query service response might need to iterate and would be easy. But, move to the map. If, requires, will move to LinkedHashMap.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
kotharironak: A step is one of the arguments in PromQL range query API - https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries

So, when we create a time series request between time windows (start, end), step is param that helps create that series. So, you will have points at (start time, value), (star time + step, value), (start time + 2*step, value), so on

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
kotharironak: No need. Broken PromQLQuery objects into PromQLInstantQuery and PromQLRangeQuery.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
```diff
@@ -0,0 +1,93 @@
...
+    @JsonProperty("result")
+    private List<PromQLMetricResult> result;
+  }
+
+  @Getter
+  @Builder
+  @AllArgsConstructor
+  @NoArgsConstructor
+  @EqualsAndHashCode
+  static class PromQLMetricResult {
```
kotharironak, line 51: Here, Value + Builder with Singular gives me an error as Value makes it final and has to initialize. So, going with other existing notations.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLInstantQuery.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.NonNull;
+import lombok.Singular;
+
+@Getter
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+class PromQLInstantQuery {
+  @NonNull @Singular private List<String> queries;
+
+  @NonNull private Instant evalTime;
```
rish691, line 18: for instant query also we need start and end time?

kotharironak: No for Instant query, just evaluation time is required, and kept the name based on the argument of PromQL API - https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries

As describe here - https://github.com/hypertrace/query-service/pull/110/#issue-1065988936, the time filter duration (endTime - startTime) will go into duration query expression of PromQL`[]`, and your endTime QS's time filter will be `evalTime`

Repeating the above example: 
```
sum by (backend_id) 
(sum_over_time(num_calls{tenant_id="__default"}[3600000ms])) 
@ 1631181508474

sum by (backend_id) 
(sum_over_time(error_count{tenant_id="__default"'}[3600000ms])) 
@ 1631181508474
```

As in the above example, see `[3600000ms]` that's your time filter of QS's aggregation query. And `@ 1631181508474` is your QS's time filter endTime, is `evalTime` for promQL query.

```
filter {
  childFilter {
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: GE
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631177908474
          }
        }
      }
    }
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: LT
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631181508474
          }
        }
      }
    }
  }
}
```

rish691: Ok, maybe you can add a comment on what eval time is?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLRangeQuery.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Singular;
+
+@Getter
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
```
rish691, line 15: why private constructor?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLInstantQuery.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Singular;
+
+@Getter
+@Builder
```
rish691, line 13: do we need builder for this class now?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLInstantQuery.java
kotharironak: Added comments.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLRangeQuery.java
kotharironak: Removed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
aaron-steinfeld: We should try to be consistent within the code - especially within a single service. I think we call this period elsewhere - can we update one or the other so the two match?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
aaron-steinfeld: Hmm - not sure I follow. What's initializing? Is it a compile error? If the two can't be used together then make sense, but we should still use `@Value` by default in other classes (seen too many bugs on missing equals and hash codes!)

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLInstantQuery.java
aaron-steinfeld: Ah, I asked this on #111 - there's quite a bit of overlap. So an instant query still applies to a range of time, but one argument is embedded in the query string and one is in a query param?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
aaron-steinfeld: OK pulled it down and looked. It's not the singular that's the issue, it's combining noargs constructor without a default value. So you've got a few options. You can use `@NoArgsConstructor(force = true)` to assign default values, or you can remove no arg constructor and use `@Jacksonized` to point jackson to the builder instead.

I'd expect you'd need something similar on the other classes too - PromQLMetricResponse and PromQLData. If I recall, by default Jackson looks for a no arg constructor which aren't provided on those classes.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLInstantQuery.java
rish691: yes, that is correct

query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientTest.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Test
+  public void testSingleInstantQuery() throws IOException {
+    mockWebServer.enqueue(getSuccessMockResponse("promql_error_count_vector.json"));
+
+    PromQLInstantQuery query =
+        PromQLInstantQuery.builder()
+            .query("errorCount")
+            .evalTime(Instant.ofEpochMilli(1435781451000L))
```
sarthak77, line 45: Is this some random time? If yes, can we declare this variable somewhere ?

sarthak77, line 45: Can we have endtime = starttime + some value ?

query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+  @Test
+  public void testSingleRangeQuery() throws IOException {
+    mockWebServer.enqueue(getSuccessMockResponse("promql_error_count_matrix.json"));
+
+    PromQLRangeQuery query =
+        PromQLRangeQuery.builder()
+            .query("errorCount")
+            .startTime(Instant.ofEpochMilli(1435781430000L))
+            .endTime(Instant.ofEpochMilli(1435781460000L))
```
sarthak77, line 62: Can we have endtime = starttime + some value ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
kotharironak: Thanks, incorporating the above change. I was not familiar with `jacksonized`.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
kotharironak: As discussed here (https://github.com/hypertrace/query-service/pull/111#discussion_r764060192), going ahead with `period` with comments.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLMetricResponse.java
kotharironak: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLQuery.java
kotharironak: Done

query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientTest.java
kotharironak: We can do that, but going with ease for corresponding API argument - https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries

rish691 (approved):

PR #113, opened 2021-12-09
feat: adds support for building converter for query service response builder from PromQL response

kotharironak:
This PR is for handling tickets - https://github.com/hypertrace/hypertrace/issues/326

This PR converts the PormQL response structure of instant query and range query to Query 
Ref: 
Instant Query Response: https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries
Range Query Response: https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries

QS Response: https://github.com/hypertrace/query-service/blob/main/query-service-api/src/main/proto/response.proto#L22

In this PR, Prepare the Row structure of Query Service Response.

As for a single QS request, it can result in multiple PromQL metric query requests. However, the response of each PromQL query request only differs on Metric name and value, the rest of the argument for attributes are the same.
So, the approach here is that we prepare the rest of the attributes (for QS Row) from one of the PromQL responses, and for metric values columns, we create columns from the individual response.

As an example:
```
filter {
  childFilter {
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: GE
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631177908474
          }
        }
      }
    }
    childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: LT
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631181508474
          }
        }
      }
    }
  }
}
selection {
  columnIdentifier {
    columnName: "BACKEND.id"
  }
}
selection {
  function {
    functionName: "SUM"
    arguments {
      columnIdentifier {
        columnName: "BACKEND.errorCount"
      }
    }
    alias: "SUM_BACKEND.errorCount_[]"
  }
}
selection {
  function {
    functionName: "SUM"
    arguments {
      columnIdentifier {
        columnName: "BACKEND.numCalls"
      }
    }
    alias: "SUM_BACKEND.numCalls_[]"
  }
}
groupBy {
  columnIdentifier {
    columnName: "BACKEND.id"
  }
}
limit: 10000
```

Corresponding PromQL queries:
```
sum by (backend_id) 
(sum_over_time(num_calls{tenant_id="__default"}[3600000ms])) 
@ 1631181508474

sum by (backend_id) 
(sum_over_time(error_count{tenant_id="__default"'}[3600000ms])) 
@ 1631181508474
```

So, for the above example, while preparing row, 
- `backend_id` information is collected form one of the response
- and num_call and error_count for the matching `backend_id` will be collected from the individual responses.


kotharironak:
@rish691 you can work on top of it for hooking the client.

kotharironak:
@rish691 added the test. It returns the result both for instant query and range query. It is also able to merge multiple metrics. However, need to optimize, but you can work on top of it.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedResponseBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+   *
+   * E.g
+   * columnNameToAttributeMap :  (SERVICE.id, service_id)
+   * columnNameToQueryMap (SERVICE.numCalls, sum by (service_id) (sum_over_time(num_calls{}[]))
+   * columnSet : [SERVICE.startTime, SERVICE.id, SERVICE.numCall]
+   * timeStampColumn : SERVICE.startTime
+   * */
+  static List<Row> buildResponse(
+      Map<Request, PromQLMetricResponse> promQLMetricResponseMap,
+      Map<String, String> columnNameToAttributeMap,
```
rish691, line 31: nit: can we rename the param names?
```
Map<String, String> logicalAttributeNameToMetricAttributeMap,
Map<String, String> logicalAttributeNameToMetricQueryMap,
List<String> columnSelectionSet,
```

kotharironak: done. All remove some of the unused code.

kotharironak:
Integration tests pass locally but fail on GHA.
Screenshot 2021-12-13 at 4 42 03 PM

query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java
```diff
@@ -76,7 +76,7 @@ public static void setup() throws Exception {
     network = Network.newNetwork();
 
     kafkaZk =
-        new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka"))
+        new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.0.0"))
```
rish691, line 79: nit: lets add a todo here to use latest version again

rish691 (approved):

PR #119, opened 2021-12-10
feat: Integrate prometheus converter, client and response builder 

rish691:
Update PrometheusBasedRequestHandler to convert queryRequest to Prometheus query, invoke the rest client to fire the query and finally convert the response to query response


query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLRangeQueries.java
```diff
@@ -2,16 +2,15 @@
...
 import lombok.Builder;
 import lombok.NonNull;
-import lombok.Singular;
 import lombok.Value;
 
 @Value
 @Builder
 class PromQLRangeQueries {
-  @NonNull @Singular List<String> queries;
+  @NonNull Map<String, String> metricNameToQueryMap;
```
kotharironak, line 13: why do we need to change it here?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -81,6 +93,35 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
...
+      PromQLRangeQueries promQLRangeQueries =
+          requestToPromqlConverter.convertToPromqlRangeQuery(
+              executionContext, originalRequest, executionContext.getAllSelections());
+      metricNameToQueryMap = promQLRangeQueries.getMetricNameToQueryMap();
+      responseMap = prometheusRestClient.executeRangeQuery(promQLRangeQueries);
+    } else {
+      PromQLInstantQueries promQLInstantQueries =
+          requestToPromqlConverter.convertToPromqlInstantQuery(
+              executionContext, originalRequest, executionContext.getAllSelections());
+      metricNameToQueryMap = promQLInstantQueries.getMetricNameToQueryMap();
```
kotharironak, line 108: If, we are going to change the `promQLInstantQueries` and `PromQLRangeQueries` with map<> internally, we will not need this input to builder.

kotharironak:
@rish691 build are failing here.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PromQLRangeQueries.java
kotharironak: we can convert this method - `buildPromqlQueries` to return `map` right now.

kotharironak:
@rish691 I have fixed the build, and if you would like to take the previous approach, you can revert back to the last commit.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestToPromqlConverter.java
```diff
@@ -148,4 +168,11 @@ private String convertColumnAttributeToString(Expression expression) {
     return prometheusViewDefinition.getPhysicalColumnNameForLogicalColumnName(
         getLogicalColumnNameForSimpleColumnExpression(expression));
   }
+
+  private String getColumnNameForMetricFunction(Expression functionExpression) {
```
kotharironak, line 172: we need to make sure that the same be used preparing columnSet.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -74,13 +91,74 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
...
+                  QueryRequestUtil.getLogicalColumnNameForSimpleColumnExpression(expression));
+              break;
+            case FUNCTION:
+              if (QueryRequestUtil.isDateTimeFunction(expression)) {
+                columnSet.add(executionContext.getTimeFilterColumn());
+              } else {
+                Function function = expression.getFunction();
+                String columnName =
+                    QueryRequestUtil.getLogicalColumnNameForSimpleColumnExpression(
+                        function.getArguments(0));
```
kotharironak, line 153: Can we have a common function for forming logical name `<function_name>:<attributeId>`?

kotharironak:
@rish691 Can you also add a description for this PR?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -3,30 +3,48 @@
...
 
-  PrometheusBasedRequestHandler(String name, Config config) {
+  PrometheusBasedRequestHandler(String name, Config config, String clientConfig) {
     this.name = name;
     this.processConfig(config);
     this.queryRequestEligibilityValidator =
         new QueryRequestEligibilityValidator(prometheusViewDefinition);
+    this.requestToPromqlConverter = new QueryRequestToPromqlConverter(prometheusViewDefinition);
+    String[] hostPort = clientConfig.split(":");
+    this.prometheusRestClient =
```
kotharironak, line 46: Can we have a singleton or have a factory way of constructing it? We can use the same RestClient across all the instance of Request Handler.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -3,30 +3,56 @@
...
     this.name = name;
     this.processConfig(config);
     this.queryRequestEligibilityValidator =
         new QueryRequestEligibilityValidator(prometheusViewDefinition);
+    this.requestToPromqlConverter = new QueryRequestToPromqlConverter(prometheusViewDefinition);
+    this.prometheusRestClient = PrometheusRestClientFactory.get().getPrometheusClient(name);
+  }
+
+  PrometheusBasedRequestHandler(
+      String name, Config config, PrometheusRestClient prometheusRestClient) {
```
kotharironak, line 49: Is this requestHanlder config? If, yes, can we rename to `requestHandlerConfig`?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -3,30 +3,56 @@
...
   PrometheusBasedRequestHandler(String name, Config config) {
     this.name = name;
     this.processConfig(config);
     this.queryRequestEligibilityValidator =
         new QueryRequestEligibilityValidator(prometheusViewDefinition);
+    this.requestToPromqlConverter = new QueryRequestToPromqlConverter(prometheusViewDefinition);
+    this.prometheusRestClient = PrometheusRestClientFactory.get().getPrometheusClient(name);
+  }
+
+  PrometheusBasedRequestHandler(
```
kotharironak, line 48: Can we mark this it to only for test or something?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -74,13 +98,70 @@
...
   public Observable<Row> handleRequest(
       QueryRequest originalRequest, ExecutionContext executionContext) {
-
     // Validate QueryContext and tenant id presence
     Preconditions.checkNotNull(executionContext);
     Preconditions.checkNotNull(executionContext.getTenantId());
 
-    // todo call convert and execute request using client here
+    Map<Request, PromQLMetricResponse> responseMap;
+    Map<String, String> metricNameToQueryMap = new LinkedHashMap<>();
```
kotharironak, line 106: metricNameToQueryMap -> logicalAttributeNameToMetricQueryMap?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -74,13 +98,70 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
...
+
+    return Observable.fromIterable(rows).doOnNext(row -> LOG.debug("collect a row: {}", row));
+  }
+
+  private boolean isRangeQueryRequest(QueryRequest queryRequest) {
+    return queryRequest.getGroupByList().stream().anyMatch(QueryRequestUtil::isDateTimeFunction);
+  }
 
-    return null;
+  private LinkedHashSet<String> prepareColumnSet(
```
kotharironak, line 140: `prepareColumnSet`-> `prepareSelectionColumnSet`

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -74,13 +98,70 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
...
+      LinkedHashSet<Expression> expressions, ExecutionContext executionContext) {
+    LinkedHashSet<String> columnSet = new LinkedHashSet<>();
```
kotharironak, line 142: columnSet -> selectionColumnSet

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
```diff
@@ -0,0 +1,43 @@
...
+  private static final PrometheusRestClientFactory INSTANCE = new PrometheusRestClientFactory();
+
+  private final ConcurrentHashMap<String, PrometheusRestClient> clientMap =
+      new ConcurrentHashMap<>();
+
+  private PrometheusRestClientFactory() {}
+
+  // Create a Prometheus Client.
+  public static PrometheusRestClient createPrometheusClient(
+      String configName, String connectionString) {
```
kotharironak, line 17: configName -> requestHandlerName?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestToPromqlConverter.java
```diff
@@ -31,33 +33,49 @@
...
       ExecutionContext executionContext,
       QueryRequest request,
-      LinkedHashSet<Expression> allSelections) {
+      LinkedHashSet<Expression> allSelections,
+      Map<String, String> metricNameToQueryMap) {
+
     QueryTimeRange queryTimeRange = getQueryTimeRange(executionContext);
-    return new PromQLInstantQueries(
+
+    metricNameToQueryMap.putAll(
```
kotharironak, line 41: metricNameToQueryMap -> logicalAttributeNameToMetricQueryMap

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
kotharironak: Can we remove this constructor?

kotharironak (approved):
overall, lgtm.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestToPromqlConverter.java
```diff
@@ -89,8 +107,10 @@ private QueryTimeRange getQueryTimeRange(ExecutionContext executionContext) {
                     && !QueryRequestUtil.isDateTimeFunction(expression))
         .map(
             functionExpression ->
-                mapToPromqlQuery(functionExpression, groupByList, filterList, duration))
-        .collect(Collectors.toUnmodifiableList());
+                ImmutablePair.of(
```
aaron-steinfeld, line 110: nit: you're packing it into a pair and then immediately unpacking it.
1. You could use an `Map.Entry` to do the same thing, which is a more appropriate fit and doesn't introduce a third party
2. In this case, you don't need to have this map call at all, since the collector is already doing a key and value mapping You could just do
```java
.collect(Collectors.toMap(
  PrometheusUtils::getColumnNameForMetricFunction,
  function -> mapToPromqlQuery(functionExpression, groupByList, filterList, duration)
));
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
```diff
@@ -14,15 +14,22 @@
...
 
-  private String host;
-  private int port;
-  private OkHttpClient okHttpClient;
+  private final String host;
+  private final int port;
+  private final OkHttpClient okHttpClient;
 
-  public PrometheusRestClient(String host, int port) {
+  PrometheusRestClient(String connection) {
```
aaron-steinfeld, line 25: Why take a connection string that we need to understand the format of? Seemed better before

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
```diff
@@ -0,0 +1,43 @@
+package org.hypertrace.core.query.service.prometheus;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+class PrometheusRestClientFactory {
+
+  // Singleton instance
+  private static final PrometheusRestClientFactory INSTANCE = new PrometheusRestClientFactory();
```
aaron-steinfeld, line 8: Aren't we already using guice?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -74,13 +88,70 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
...
+  }
+
+  private boolean isRangeQueryRequest(QueryRequest queryRequest) {
+    return queryRequest.getGroupByList().stream().anyMatch(QueryRequestUtil::isDateTimeFunction);
+  }
 
-    return null;
+  private LinkedHashSet<String> prepareSelectionColumnSet(
+      LinkedHashSet<Expression> expressions, ExecutionContext executionContext) {
+    LinkedHashSet<String> selectionColumnSet = new LinkedHashSet<>();
```
aaron-steinfeld, line 132: collect and make immutable rather than returning an immutable type. Also, why are we returning `LinkedHashSet` If we're concerned about order and need to declare it, return a list; if we're not concerned about order or we are in this function but not elsewhere, return a `Set`. In other words, does this api need to _both_ declare that it's unique and ordered, or are those implementation details?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusUtils.java
```diff
@@ -0,0 +1,14 @@
...
+import org.hypertrace.core.query.service.QueryRequestUtil;
+import org.hypertrace.core.query.service.api.Expression;
+
+class PrometheusUtils {
+  static String getColumnNameForMetricFunction(Expression functionExpression) {
+    String functionName = functionExpression.getFunction().getFunctionName().toUpperCase();
+    String columnName =
+        QueryRequestUtil.getLogicalColumnNameForSimpleColumnExpression(
+            functionExpression.getFunction().getArguments(0));
+    return String.join(":", functionName, columnName);
```
aaron-steinfeld, line 12: This makes me nervous - a lot of the work on the attribute expressions feature has been in fixing similar code where we generate aliases or use attribute IDs as column names in the result metadata and ignore the requested alias. Since this is ignoring the requested alias, is it for internal use only or does this show up in the response?

kotharironak: It's for internal mapping and internal use only (It doesn't show up in response). As we have to fire multiple queries for the same QS query, and if that query has two metrics function on the same attributed, we needed a unique mapping.

e.g 
```
select SERVICE.service_id, sum(SERVICE.numCalls), avg(SERVICE.numCalls).
```

As in the above example, the same attribute id is used for two different metric functions. It wasn't the issue in the case of PinotBase response where it can be mapped using column index. But, here, we will have two PromQL responses - one for `sum(SERVICE.numCalls)` and one for `avg(SERVICE.numCalls)`. So, internally, we need a unique map of the response, so we are using `<funcname>:<attributeId>` internally for response building.

aaron-steinfeld: Got it; In other places, we were building the response metadata like this, rather than using the alias. As long as it's switching back to that before returning, we're good.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
rish691: The requirement from this method is that the returned collection has unique element and the order is same as the input collection. 

Now returning set from this method would mean losing the order. 
A list can be returned as it will maintain the order (the input collection already has unique elements so will the list)

If you see the flow of data,
this method is invoked on a collection in the ExecutionContext class, which returns object of type LinkedHashSet, 
then that collection is processed by this method. Finally it is passed into the response builder class a method param. 
```
PrometheusBasedResponseBuilder.buildResponse(
            responseMap,
            prometheusViewDefinition.getAttributeMap(),
            logicalAttributeNameToMetricQueryMap,
            prepareSelectionColumnSet(executionContext.getAllSelections(), executionContext),
            executionContext.getTimeFilterColumn());
```

Checking the users of `executionContext.getAllSelections()`, I think this collection itself could have been of `List` type

I understand that wherever possible we should return interface, but how to convey that the param `LinkedHashSet<String> columnSelectionSet` (which will be changed to List<String> columnSelectionSet) in this method org.hypertrace.core.query.service.prometheus.PrometheusBasedResponseBuilder#buildResponse must have ordered and unique elements?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
rish691: This is driven from the config itself, where we are taking a string called `connectionString`, so just wanted to stick with that in code as well.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestToPromqlConverter.java
rish691: yeah of course, not sure how this code got in

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
rish691: Tried using it, isn't looking very clean

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
aaron-steinfeld: Isn't that our config or am I misunderstanding? Why not just take them as a distinct host and port like we do everywhere else?

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
aaron-steinfeld: > I understand that wherever possible we should return interface, but how to convey that the param LinkedHashSet<String> columnSelectionSet (which will be changed to List<String> columnSelectionSet) in this method org.hypertrace.core.query.service.prometheus.PrometheusBasedResponseBuilder#buildResponse must have ordered and unique elements?

The problem here isn't just "use an interface instead of an implementation", it's what that rule of thumb is based on - that returning the type is leaking implementation details. It's much less about the implementation of the collection itself, it's that we're trying to convey the responsibilities of one class to another with it. We only need (or want) one class to understand how to prepare selections, and that's the only place that needs to know if duplicates are allowed or not. The caller can then trust that prepare the selection column set was responsible for doing everything involved in preparing a selection set, but it doesn't need to know what that was.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
aaron-steinfeld: Hmm - IMO, the current form where we implement a singleton by hand, use statics, call a static method and ignore its return to modify state and then needlessly take in a handler name is significantly harder to understand and objectively more code. There's potential bugs in there too, because if I give a handler name that's already been registered, my connection string is ignored.


```diff
diff --git a/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java b/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
index 9f76633..355bbf3 100644
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
@@ -36,13 +36,14 @@ public class PrometheusBasedRequestHandler implements RequestHandler {
   private Optional<String> startTimeAttributeName;
   private PrometheusViewDefinition prometheusViewDefinition;
 
-  PrometheusBasedRequestHandler(String name, Config requestHandlerConfig) {
+  PrometheusBasedRequestHandler(
+      String name, Config requestHandlerConfig, PrometheusRestClient prometheusRestClient) {
     this.name = name;
     this.processConfig(requestHandlerConfig);
     this.queryRequestEligibilityValidator =
         new QueryRequestEligibilityValidator(prometheusViewDefinition);
     this.requestToPromqlConverter = new QueryRequestToPromqlConverter(prometheusViewDefinition);
-    this.prometheusRestClient = PrometheusRestClientFactory.get().getPrometheusClient(name);
+    this.prometheusRestClient = prometheusRestClient;
   }
 
   @Override
diff --git a/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRequestHandlerBuilder.java b/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRequestHandlerBuilder.java
index 4136015..acd0be7 100644
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRequestHandlerBuilder.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRequestHandlerBuilder.java
@@ -10,10 +10,14 @@ import org.hypertrace.core.query.service.RequestHandlerClientConfigRegistry;
 public class PrometheusRequestHandlerBuilder implements RequestHandlerBuilder {
 
   private final RequestHandlerClientConfigRegistry clientConfigRegistry;
+  private final PrometheusRestClientFactory prometheusRestClientFactory;
 
   @Inject
-  PrometheusRequestHandlerBuilder(RequestHandlerClientConfigRegistry clientConfigRegistry) {
+  PrometheusRequestHandlerBuilder(
+      RequestHandlerClientConfigRegistry clientConfigRegistry,
+      PrometheusRestClientFactory prometheusRestClientFactory) {
     this.clientConfigRegistry = clientConfigRegistry;
+    this.prometheusRestClientFactory = prometheusRestClientFactory;
   }
 
   @Override
@@ -32,9 +36,9 @@ public class PrometheusRequestHandlerBuilder implements RequestHandlerBuilder {
                     new UnsupportedOperationException(
                         "Client config requested but not registered: " + config.getClientConfig()));
 
-    PrometheusRestClientFactory.createPrometheusClient(
-        config.getName(), clientConfig.getConnectionString());
-
-    return new PrometheusBasedRequestHandler(config.getName(), config.getRequestHandlerInfo());
+    return new PrometheusBasedRequestHandler(
+        config.getName(),
+        config.getRequestHandlerInfo(),
+        prometheusRestClientFactory.getPrometheusClient(clientConfig.getConnectionString()));
   }
 }
diff --git a/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java b/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
index a1b7bad..e1c8d60 100644
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClientFactory.java
@@ -1,43 +1,15 @@
 package org.hypertrace.core.query.service.prometheus;
 
 import java.util.concurrent.ConcurrentHashMap;
+import javax.inject.Singleton;
 
+@Singleton
 class PrometheusRestClientFactory {
 
-  // Singleton instance
-  private static final PrometheusRestClientFactory INSTANCE = new PrometheusRestClientFactory();
-
   private final ConcurrentHashMap<String, PrometheusRestClient> clientMap =
       new ConcurrentHashMap<>();
 
-  private PrometheusRestClientFactory() {}
-
-  // Create a Prometheus Client.
-  public static PrometheusRestClient createPrometheusClient(
-      String requestHandlerName, String connectionString) {
-    if (!get().containsClient(requestHandlerName)) {
-      synchronized (get()) {
-        if (!get().containsClient(requestHandlerName)) {
-          get().addPrometheusClient(requestHandlerName, new PrometheusRestClient(connectionString));
-        }
-      }
-    }
-    return get().getPrometheusClient(requestHandlerName);
-  }
-
-  public static PrometheusRestClientFactory get() {
-    return INSTANCE;
-  }
-
-  private void addPrometheusClient(String requestHandlerName, PrometheusRestClient client) {
-    this.clientMap.put(requestHandlerName, client);
-  }
-
-  public boolean containsClient(String requestHandlerName) {
-    return this.clientMap.containsKey(requestHandlerName);
-  }
-
-  public PrometheusRestClient getPrometheusClient(String requestHandlerName) {
-    return this.clientMap.get(requestHandlerName);
+  public PrometheusRestClient getPrometheusClient(String connectionString) {
+    return clientMap.computeIfAbsent(connectionString, PrometheusRestClient::new);
   }
 }
diff --git a/query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandlerTest.java b/query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandlerTest.java
index 26a4607..b4d4c09 100644
--- a/query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandlerTest.java
+++ b/query-service-impl/src/test/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandlerTest.java
@@ -40,11 +40,12 @@ class PrometheusBasedRequestHandlerTest {
     mockWebServer.start(9099);
 
     Config config = PrometheusTestUtils.getDefaultPrometheusConfig();
-    PrometheusRestClientFactory.createPrometheusClient(
-        config.getString(CONFIG_PATH_NAME), config.getString(CONFIG_PATH_CLIENT_KEY));
+
     prometheusBasedRequestHandler =
         new PrometheusBasedRequestHandler(
-            config.getString(CONFIG_PATH_NAME), config.getConfig(CONFIG_PATH_REQUEST_HANDLER_INFO));
+            config.getString(CONFIG_PATH_NAME),
+            config.getConfig(CONFIG_PATH_REQUEST_HANDLER_INFO),
+            new PrometheusRestClient(config.getString(CONFIG_PATH_CLIENT_KEY)));
   }
 
   @AfterAll
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
rish691: > Why not just take them as a distinct host and port like we do everywhere else

We don't do it that way in here. 

When the app config is read, the client config is represented by `RequestHandlerClientConfig` inner class from https://github.com/hypertrace/query-service/blob/main/query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java#L78, 

```
  public static class RequestHandlerClientConfig {
    private static final String CONFIG_PATH_TYPE = "type";
    private static final String CONFIG_PATH_CONNECTION_STRING = "connectionString";
    private String type;
    private String connectionString;
```

Now I could have modified it to add host and port additionally and only use them for prom, but the current approach felt better to me. What do you say?

aaron-steinfeld: OK got it - the pinot version uses a connection string (because it uses zookeeper and is not just a host/port pair), and the way the handler config is architected, all handlers regardless of type share config classes. I think the shared code between different handler types is problematic (for reasons like this), but that's a much broader problem and far out of scope.

Resolving.

PR #123, opened 2021-12-14
Metrics followup 

rish691:
From https://github.com/hypertrace/query-service/pull/119


query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestToPromqlConverter.java
```diff
@@ -105,12 +104,10 @@ private QueryTimeRange getQueryTimeRange(ExecutionContext executionContext) {
             expression ->
                 expression.getValueCase().equals(ValueCase.FUNCTION)
                     && !QueryRequestUtil.isDateTimeFunction(expression))
-        .map(
-            functionExpression ->
-                ImmutablePair.of(
-                    PrometheusUtils.getColumnNameForMetricFunction(functionExpression),
-                    mapToPromqlQuery(functionExpression, groupByList, filterList, duration)))
-        .collect(Collectors.toMap(ImmutablePair::getLeft, ImmutablePair::getRight));
+        .collect(
```
kotharironak, line 107: Ah! we missed this during the review. This was the case before we added this function `getColumnNameForMetricFunction` of fixing the same attribute with two different metric aggregation function situations - `select SERVICE.service_id, sum(SERVICE.numCalls), sum(SERVICE.errorCount)`

If ImmputablePair is not required anywhere, you can remove its lib deps from build too, can you check if it's used anywhere else?

kotharironak:
@rish691 As part of this PR, can you put the constraints on log4j related library coming from a third party to use the latest `2.15.0` as part of snyk failures.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
```diff
@@ -125,12 +123,12 @@ public OkHttpResponseCallback(Request request) {
     }
 
     @Override
-    public void onResponse(Call call, Response response) throws IOException {
+    public void onResponse(@NonNull Call call, @NonNull Response response) {
```
kotharironak, line 126: why do we need explicitly `@NonNull`?

query-service-impl/build.gradle.kts
```diff
@@ -49,7 +49,6 @@ dependencies {
   implementation("com.squareup.okhttp3:okhttp:4.9.1")
   annotationProcessor("org.projectlombok:lombok:1.18.20")
   compileOnly("org.projectlombok:lombok:1.18.20")
-  implementation("org.apache.commons:commons-lang3:3.10")
```
aaron-steinfeld, line 52: 🥳

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -36,13 +36,16 @@
...
+      String name,
+      Config requestHandlerConfig,
+      PrometheusRestClientFactory prometheusRestClientFactory) {
     this.name = name;
     this.processConfig(requestHandlerConfig);
     this.queryRequestEligibilityValidator =
         new QueryRequestEligibilityValidator(prometheusViewDefinition);
     this.requestToPromqlConverter = new QueryRequestToPromqlConverter(prometheusViewDefinition);
-    this.prometheusRestClient = PrometheusRestClientFactory.get().getPrometheusClient(name);
+    this.prometheusRestClient = prometheusRestClientFactory.getPrometheusClient(name);
```
aaron-steinfeld, line 48: Was responding to the other one before reviewing this - see https://github.com/hypertrace/query-service/pull/119#discussion_r768704268 (the two part registration/lookup isn't needed)

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestToPromqlConverter.java
aaron-steinfeld: So `getColumnNameForMetricFunction` returns a unique value for `sum(SERVICE.numCalls)` and `sum(SERVICE.errorCount)`? If that's the case, it's not really the column name (which is `SERVICE.numCalls` in both cases), so maybe `generateAliasForMetricFunction` might be more appropriate. If it's not the case, this code would error in the above example.

rish691: done

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusRestClient.java
rish691: IDE suggestion, it says `Not annotated parameter overrides @NotNull parameter `, basically this param is marked not null in the parent

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/PrometheusBasedRequestHandler.java
```diff
@@ -127,30 +129,26 @@ private boolean isRangeQueryRequest(QueryRequest queryRequest) {
...
-            case FUNCTION:
-              if (QueryRequestUtil.isDateTimeFunction(expression)) {
-                selectionColumnSet.add(executionContext.getTimeFilterColumn());
-              } else {
-                selectionColumnSet.add(PrometheusUtils.getColumnNameForMetricFunction(expression));
+    return expressions.stream()
+        .map(
+            expression -> {
+              ValueCase valueCase = expression.getValueCase();
+              switch (valueCase) {
```
aaron-steinfeld, line 138: nit
```suggestion
              switch (expression.getValueCase()) {
```

aaron-steinfeld (approved):

Issue #122, opened 2021-12-14
Add support for `CONTAINS_KEY` and `NOT_CONTAINS_KEY` for attributeExpression


sarthak77:
- [x] Add support for simple attribute expression in `CONTAINS_KEY`
- [x] Add support for `NOT_CONTAINS_KEY`

Issue #109, opened 2021-11-29
Add support for filtering map attributes using regex operator


sarthak77:
Based on [this](https://github.com/hypertrace/query-service/pull/107#discussion_r758037652) discussion, add support for filtering using regex operator. 

eg. Working pinot query : 

```
select * from spanEventView
where start_time_millis > 1628187491000 and start_time_millis < 1638188391000
and tags__KEYS = 'span.kind'
and mapValue(tags__KEYS, 'span.kind', tags__VALUES) != ''  
and REGEXP_LIKE(mapValue(tags__KEYS, 'span.kind', tags__VALUES), '^client*')
limit 10
```

PR #125, opened 2021-12-15
Fix : Add support for contains key and like operator

sarthak77:
Ref issue : https://github.com/hypertrace/query-service/issues/122 , https://github.com/hypertrace/query-service/issues/109

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -340,7 +340,7 @@ class QueryRequestToPinotSQLConverter {
       }
     }
 
-    List<LiteralConstant> literalConstantList = new ArrayList<>(List.of());
+    List<LiteralConstant> literalConstantList = new ArrayList<>();
     for (int i = 0; i < 2; i++) {
       literalConstantList.add(
           LiteralConstant.newBuilder()
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -9,6 +9,7 @@ import static org.hypertrace.core.query.service.api.Expression.ValueCase.LITERAL
 import com.google.common.base.Joiner;
 import com.google.protobuf.ByteString;
 import java.util.AbstractMap.SimpleEntry;
+import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map.Entry;
@@ -155,10 +156,10 @@ class QueryRequestToPinotSQLConverter {
           builder.append(")");
           break;
         case CONTAINS_KEY:
-          LiteralConstant[] kvp = convertExpressionToMapLiterals(filter.getRhs());
+          List<LiteralConstant> kvp = convertExpressionToMapLiterals(filter.getRhs());
           builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
           builder.append(" = ");
-          builder.append(convertLiteralToString(kvp[MAP_KEY_INDEX], paramsBuilder));
+          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
           break;
         case CONTAINS_KEYVALUE:
           kvp = convertExpressionToMapLiterals(filter.getRhs());
@@ -166,21 +167,21 @@ class QueryRequestToPinotSQLConverter {
           String valCol = convertExpressionToMapValueColumn(filter.getLhs());
           builder.append(keyCol);
           builder.append(" = ");
-          builder.append(convertLiteralToString(kvp[MAP_KEY_INDEX], paramsBuilder));
+          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
           builder.append(" AND ");
           builder.append(valCol);
           builder.append(" = ");
-          builder.append(convertLiteralToString(kvp[MAP_VALUE_INDEX], paramsBuilder));
+          builder.append(convertLiteralToString(kvp.get(MAP_VALUE_INDEX), paramsBuilder));
           builder.append(" AND ");
           builder.append(MAP_VALUE);
           builder.append("(");
           builder.append(keyCol);
           builder.append(",");
-          builder.append(convertLiteralToString(kvp[MAP_KEY_INDEX], paramsBuilder));
+          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
           builder.append(",");
           builder.append(valCol);
           builder.append(") = ");
-          builder.append(convertLiteralToString(kvp[MAP_VALUE_INDEX], paramsBuilder));
+          builder.append(convertLiteralToString(kvp.get(MAP_VALUE_INDEX), paramsBuilder));
           break;
         default:
           rhs = handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
@@ -323,34 +324,31 @@ class QueryRequestToPinotSQLConverter {
     throw new IllegalArgumentException("operator supports multi value column only");
   }
 
-  private LiteralConstant[] convertExpressionToMapLiterals(Expression expression) {
-    LiteralConstant[] literals = new LiteralConstant[2];
+  private List<LiteralConstant> convertExpressionToMapLiterals(Expression expression) {
+    List<String> literals = new ArrayList<>(List.of("", ""));
     if (expression.getValueCase() == LITERAL) {
       LiteralConstant value = expression.getLiteral();
-      if (value.getValue().getValueType() == ValueType.STRING_ARRAY) {
-        for (int i = 0; i < 2 && i < value.getValue().getStringArrayCount(); i++) {
-          literals[i] =
-              LiteralConstant.newBuilder()
-                  .setValue(
-                      Value.newBuilder().setString(value.getValue().getStringArray(i)).build())
-                  .build();
-        }
+      if (value.getValue().getValueType() == ValueType.STRING_ARRAY
+          && value.getValue().getStringArrayCount() == 2) {
+        literals.set(0, value.getValue().getStringArray(0));
+        literals.set(1, value.getValue().getStringArray(1));
+      } else if (value.getValue().getValueType() == ValueType.STRING) {
+        literals.set(0, value.getValue().getString());
       } else {
         throw new IllegalArgumentException(
-            "operator CONTAINS_KEYVALUE supports "
-                + ValueType.STRING_ARRAY.name()
-                + " value type only");
+            "Unsupported arguments for CONTAINS_KEY / CONTAINS_KEYVALUE operator");
       }
     }
 
-    for (int i = 0; i < literals.length; i++) {
-      if (literals[i] == null) {
-        literals[i] =
-            LiteralConstant.newBuilder().setValue(Value.newBuilder().setString("").build()).build();
-      }
+    List<LiteralConstant> literalConstantList = new ArrayList<>(List.of());
+    for (int i = 0; i < 2; i++) {
+      literalConstantList.add(
+          LiteralConstant.newBuilder()
+              .setValue(Value.newBuilder().setString(literals.get(i)).build())
+              .build());
     }
 
-    return literals;
+    return literalConstantList;
   }
 
   /** TODO:Handle all types */
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -155,16 +155,13 @@ class QueryRequestToPinotSQLConverter {
           builder.append(convertExpressionToString(rhs, paramsBuilder, executionContext));
           builder.append(")");
           break;
+        case CONTAINS_KEY:
         case NOT_CONTAINS_KEY:
           List<LiteralConstant> kvp = convertExpressionToMapLiterals(filter.getRhs());
           builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
-          builder.append(" != ");
-          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
-          break;
-        case CONTAINS_KEY:
-          kvp = convertExpressionToMapLiterals(filter.getRhs());
-          builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
-          builder.append(" = ");
+          builder.append(" ");
+          builder.append(operator);
+          builder.append(" ");
           builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
           break;
         case CONTAINS_KEYVALUE:
@@ -240,8 +237,10 @@ class QueryRequestToPinotSQLConverter {
       case NOT:
         return "NOT";
       case EQ:
+      case CONTAINS_KEY:
         return "=";
       case NEQ:
+      case NOT_CONTAINS_KEY:
         return "!=";
       case IN:
         return "IN";
@@ -257,8 +256,6 @@ class QueryRequestToPinotSQLConverter {
         return "<=";
       case LIKE:
         return REGEX_OPERATOR;
-      case NOT_CONTAINS_KEY:
-      case CONTAINS_KEY:
       case CONTAINS_KEYVALUE:
         return MAP_VALUE;
       case RANGE:
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -155,8 +155,14 @@ class QueryRequestToPinotSQLConverter {
           builder.append(convertExpressionToString(rhs, paramsBuilder, executionContext));
           builder.append(")");
           break;
-        case CONTAINS_KEY:
+        case NOT_CONTAINS_KEY:
           List<LiteralConstant> kvp = convertExpressionToMapLiterals(filter.getRhs());
+          builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
+          builder.append(" != ");
+          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
+          break;
+        case CONTAINS_KEY:
+          kvp = convertExpressionToMapLiterals(filter.getRhs());
           builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
           builder.append(" = ");
           builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
@@ -251,6 +257,7 @@ class QueryRequestToPinotSQLConverter {
         return "<=";
       case LIKE:
         return REGEX_OPERATOR;
+      case NOT_CONTAINS_KEY:
       case CONTAINS_KEY:
       case CONTAINS_KEYVALUE:
         return MAP_VALUE;
--- a/query-service-api/src/main/proto/request.proto
+++ b/query-service-api/src/main/proto/request.proto
@@ -55,6 +55,7 @@ enum Operator {
   LIKE = 12;
   CONTAINS_KEY = 13;
   CONTAINS_KEYVALUE = 14;
+  NOT_CONTAINS_KEY = 15;
 }
 
 
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryRequestUtil.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryRequestUtil.java
@@ -3,7 +3,6 @@ package org.hypertrace.core.query.service;
 import static org.hypertrace.core.query.service.api.Expression.ValueCase.ATTRIBUTE_EXPRESSION;
 import static org.hypertrace.core.query.service.api.Expression.ValueCase.COLUMNIDENTIFIER;
 
-import java.util.List;
 import org.hypertrace.core.query.service.api.AttributeExpression;
 import org.hypertrace.core.query.service.api.ColumnIdentifier;
 import org.hypertrace.core.query.service.api.Expression;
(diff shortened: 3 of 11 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -323,34 +324,31 @@ private String convertExpressionToMapValueColumn(Expression expression) {
...
-              LiteralConstant.newBuilder()
-                  .setValue(
-                      Value.newBuilder().setString(value.getValue().getStringArray(i)).build())
-                  .build();
-        }
+      if (value.getValue().getValueType() == ValueType.STRING_ARRAY
+          && value.getValue().getStringArrayCount() == 2) {
+        literals.set(0, value.getValue().getStringArray(0));
+        literals.set(1, value.getValue().getStringArray(1));
+      } else if (value.getValue().getValueType() == ValueType.STRING) {
```
aaron-steinfeld, line 335: So support for a string arg is the only functional change in this PR, right?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -323,34 +324,31 @@ private String convertExpressionToMapValueColumn(Expression expression) {
     throw new IllegalArgumentException("operator supports multi value column only");
   }
 
-  private LiteralConstant[] convertExpressionToMapLiterals(Expression expression) {
-    LiteralConstant[] literals = new LiteralConstant[2];
+  private List<LiteralConstant> convertExpressionToMapLiterals(Expression expression) {
+    List<String> literals = new ArrayList<>(List.of("", ""));
```
aaron-steinfeld, line 328: If we're going to clean up the implementation, let's do it right and return an object rather than a list.
```
@Value
class MapKeyValueArguments {
  LiteralConstant key;
  Optional<LiteralConstant> value;
}
```

rish691: can `Map.Entry<LiteralConstant, Optional<LiteralConstant>>` be used as tuple?

aaron-steinfeld: I'm a bit on the fence. I probably wouldn't do it, but I wouldn't call it out on review. In general, we'd rather create a new class than use an Entry or Pair class, but in this case it's reasonable to use Entry since we're representing a key and value (Pair pretty much never makes sense, because left and right have no meaning). The reason I'd still lean against it is that we're kind of mixing a structural paradigm with an object-oriented one. What we're producing here "is a" map entry, but it would never be used in a map - we're just using an existing unrelated class because it has the structure we're looking for.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: Mostly nit changes till now. I thought like and contains_key will not work but with the new changes, they were working so just added unit tests for them till now.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: If the key is an index like 0 or 1, then why do we need it to be a literalConstant ? Then at the places where we access using 0/1, we would need to create a literalConstant first.

sarthak77: Currently, we use a list. Here we do `x.get(map_key_index) where map_key_index is 0`. 
If we go by the suggested approach, then we would have to create a literal constant for 0/1. Since the key is fairly simple (only 0/1), can we use something like this ?
```
@Value
class MapKeyValueArguments {
  int key;
  Optional<LiteralConstant> value;
}
```

kotharironak: As a second thought (after looking at the test of the LIKE operator with attribute expression), should we handle attribute expression for `CONTAINS_KEY` and `NOT_CONTAINS_KEY` as part of this PR? 

Is there any other place, we have not still handled `attribute expression`?

sarthak77: Everything is covered now I guess.

kotharironak: So, as part of this PR, shall we cover `CONTAINS_KEY and NOT_CONTAINS_KEY` for attribute expression as well?

kotharironak: > Currently, we use a list. Here we do x.get(map_key_index) where map_key_index is 0.
If we go by the suggested approach, then we would have to create a literal constant for 0/1. Since the key is fairly simple (only 0/1), can we use something like this ?

How does the `int key` value be used? How will we get LiteralConsant, will we still have List<LiteralConsant>? Based on aaron's suggestion, you can do something as below. And, you can access `key` literal constant directly from the object.

```
diff --git a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/MapKeyValueArguments.java b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/MapKeyValueArguments.java
index 84d627a..89ecba1 100644
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/MapKeyValueArguments.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/MapKeyValueArguments.java
@@ -1,2 +1,11 @@
-package org.hypertrace.core.query.service.pinot;public class MapKeyValueArguments {
+package org.hypertrace.core.query.service.pinot;
+
+import java.util.Optional;
+import lombok.Value;
+import org.hypertrace.core.query.service.api.LiteralConstant;
+
+@Value
+class MapKeyValueArguments {
+  LiteralConstant key;
+  Optional<LiteralConstant> value;
 }
diff --git a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
index 557a670..e99f1eb 100644
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -9,10 +9,10 @@ import static org.hypertrace.core.query.service.api.Expression.ValueCase.LITERAL
 import com.google.common.base.Joiner;
 import com.google.protobuf.ByteString;
 import java.util.AbstractMap.SimpleEntry;
-import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.Optional;
 import org.hypertrace.core.query.service.ExecutionContext;
 import org.hypertrace.core.query.service.api.Expression;
 import org.hypertrace.core.query.service.api.Filter;
@@ -156,38 +156,41 @@ class QueryRequestToPinotSQLConverter {
           builder.append(")");
           break;
         case NOT_CONTAINS_KEY:
-          List<LiteralConstant> kvp = convertExpressionToMapLiterals(filter.getRhs());
+          MapKeyValueArguments mapKeyValueArguments =
+              convertExpressionToMapLiterals(filter.getRhs());
           builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
           builder.append(" != ");
-          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
+          builder.append(convertLiteralToString(mapKeyValueArguments.getKey(), paramsBuilder));
           break;
         case CONTAINS_KEY:
-          kvp = convertExpressionToMapLiterals(filter.getRhs());
+          mapKeyValueArguments = convertExpressionToMapLiterals(filter.getRhs());
           builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
           builder.append(" = ");
-          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
+          builder.append(convertLiteralToString(mapKeyValueArguments.getKey(), paramsBuilder));
           break;
         case CONTAINS_KEYVALUE:
-          kvp = convertExpressionToMapLiterals(filter.getRhs());
+          mapKeyValueArguments = convertExpressionToMapLiterals(filter.getRhs());
           String keyCol = convertExpressionToMapKeyColumn(filter.getLhs());
           String valCol = convertExpressionToMapValueColumn(filter.getLhs());
           builder.append(keyCol);
           builder.append(" = ");
-          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
+          builder.append(convertLiteralToString(mapKeyValueArguments.getKey(), paramsBuilder));
           builder.append(" AND ");
           builder.append(valCol);
           builder.append(" = ");
-          builder.append(convertLiteralToString(kvp.get(MAP_VALUE_INDEX), paramsBuilder));
+          builder.append(
+              convertLiteralToString(mapKeyValueArguments.getValue().get(), paramsBuilder));
           builder.append(" AND ");
           builder.append(MAP_VALUE);
           builder.append("(");
           builder.append(keyCol);
           builder.append(",");
-          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
+          builder.append(convertLiteralToString(mapKeyValueArguments.getKey(), paramsBuilder));
           builder.append(",");
           builder.append(valCol);
           builder.append(") = ");
-          builder.append(convertLiteralToString(kvp.get(MAP_VALUE_INDEX), paramsBuilder));
+          builder.append(
+              convertLiteralToString(mapKeyValueArguments.getValue().get(), paramsBuilder));
           break;
         default:
           rhs = handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
@@ -331,31 +334,37 @@ class QueryRequestToPinotSQLConverter {
     throw new IllegalArgumentException("operator supports multi value column only");
   }
 
-  private List<LiteralConstant> convertExpressionToMapLiterals(Expression expression) {
-    List<String> literals = new ArrayList<>(List.of("", ""));
+  private MapKeyValueArguments convertExpressionToMapLiterals(Expression expression) {
+    LiteralConstant mapKey = null;
+    Optional<LiteralConstant> mapValue = Optional.empty();
+
     if (expression.getValueCase() == LITERAL) {
       LiteralConstant value = expression.getLiteral();
       if (value.getValue().getValueType() == ValueType.STRING_ARRAY
           && value.getValue().getStringArrayCount() == 2) {
-        literals.set(0, value.getValue().getStringArray(0));
-        literals.set(1, value.getValue().getStringArray(1));
+        mapKey =
+            LiteralConstant.newBuilder()
+                .setValue(Value.newBuilder().setString(value.getValue().getStringArray(0)).build())
+                .build();
+
+        mapValue =
+            Optional.of(
+                LiteralConstant.newBuilder()
+                    .setValue(
+                        Value.newBuilder().setString(value.getValue().getStringArray(1)).build())
+                    .build());
       } else if (value.getValue().getValueType() == ValueType.STRING) {
-        literals.set(0, value.getValue().getString());
+        mapKey =
+            LiteralConstant.newBuilder()
+                .setValue(Value.newBuilder().setString(value.getValue().getString()).build())
+                .build();
       } else {
         throw new IllegalArgumentException(
             "Unsupported arguments for CONTAINS_KEY / CONTAINS_KEYVALUE operator");
       }
     }
 
-    List<LiteralConstant> literalConstantList = new ArrayList<>();
-    for (int i = 0; i < 2; i++) {
-      literalConstantList.add(
-          LiteralConstant.newBuilder()
-              .setValue(Value.newBuilder().setString(literals.get(i)).build())
-              .build());
-    }
-
-    return literalConstantList;
+    return new MapKeyValueArguments(mapKey, mapValue);
   }
 
   /** TODO:Handle all types */

```

sarthak77: >So, as part of this PR, shall we cover CONTAINS_KEY and NOT_CONTAINS_KEY for attribute expression as well? Like in this case convertExpressionToMapLiterals which returns key in index 0. For attribute expression, it needs to fetch from its subPath, right?

Then also it will still have rhs right ?
Like we can say `Span.tags CONTAINS_KEY 'flags'` or `lhs CONTAINS_KEY rhs`
For attribute expression how will we use the rhs expression ?

Currently we have these function : 
```
  public static Filter createContainsKeyFilter(AttributeExpression complexAttributeExpression) {
    return createContainsKeyFilter(
        complexAttributeExpression.getAttributeId(), complexAttributeExpression.getSubpath());
  }

  public static Filter createContainsKeyFilter(String column, String value) {
    return Filter.newBuilder()
        .setOperator(Operator.CONTAINS_KEY)
        .setLhs(createSimpleAttributeExpression(column))
        .setRhs(createStringLiteralValueExpression(value))
        .build();
  }
```

sarthak77: >How does the int key value be used? How will we get LiteralConsant, will we still have List? Based on aaron's suggestion, you can do something as below. And, you can access key literal constant directly from the object. e.g builder.append(convertLiteralToString(mapKeyValueArguments.getKey(), paramsBuilder));

Ok got it now. Earlier I thought 0 is key. But you are proposing that key should be `Span.tags` and value should be say `span.kind`. 

Should I make this change then ?

kotharironak: > Then also it will still have rhs right ?
Like we can say Span.tags CONTAINS_KEY 'flags' or lhs CONTAINS_KEY rhs

Yes, you are right. Rhs is a key in map. I got confused.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -323,34 +331,31 @@ private String convertExpressionToMapValueColumn(Expression expression) {
...
     if (expression.getValueCase() == LITERAL) {
```
sarthak77, line 336: Also, do we throw an exception here if rhs is not literal ? As using the below design, we always expect the key. Also, we are always calling this method from such a case so it may prevent any error if some new case is added. 

```
@Value class MapKeyValueArguments { LiteralConstant key; Optional<LiteralConstant> value; }
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
kotharironak: >> How does the int key value be used? How will we get LiteralConsant, will we still have List? Based on aaron's suggestion, you can do something as below. And, you can access key literal constant directly from the object. e.g builder.append(convertLiteralToString(mapKeyValueArguments.getKey(), paramsBuilder));

> Ok got it now. Earlier I thought 0 is key. But you are proposing that key should be Span.tags and value should be say span.kind.

For a map, there are two arguments one is key and the other is value - e.g x['key'] = 'value'
So, here `Span.tags` is a map, the `span.kind` is a key arg. 
In the case of `CONTAINS_KEY` and `NOT_CONTAINS_KEY`, it will have only one `key` arg. But, in the case of `CONTAINS_KEY_VALUE`, it will have both `key` and `value` args, and so `value` arg is optional in structure.
e.g `Span.tags['span.kind'] = 'client'`, for this `key` will be `span.kind` and `value` will be `client`

> Should I make this change then ?
Yes, we can do that.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: CC @kotharironak

kotharironak: In the existing `main` code, we are defaulting to empty string `""` if the case is not `LITERAL`. Here as well default to `""` string LiteralConstant.

sarthak77: I thought optional was to deal with this default ""

kotharironak: Optional<LiteralConstant> value means that the map's value argument is not mandatory. Refer to this - https://github.com/hypertrace/query-service/pull/125#discussion_r770359421

kotharironak: So, in a way, we can also extend as you suggested that we expect the `key` argument should be present. if it's not there, we can throw the exception.

Does the below make sense?
`spanTags contains_key ''`

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
aaron-steinfeld: Alternatively, we can wait then drop this whole thing. Assuming we deprecate CONTAINS_KEY_VALUE and remove the array around the contains_key arg, this whole function can just go away I think?

sarthak77: Yes. We can simplify that and move inline maybe.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -154,33 +155,39 @@ private String convertFilterToString(
           builder.append(convertExpressionToString(rhs, paramsBuilder, executionContext));
           builder.append(")");
           break;
+        case NOT_CONTAINS_KEY:
+          List<LiteralConstant> kvp = convertExpressionToMapLiterals(filter.getRhs());
+          builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
+          builder.append(" != ");
+          builder.append(convertLiteralToString(kvp.get(MAP_KEY_INDEX), paramsBuilder));
```
aaron-steinfeld, line 162: CONTAINS_KEY and NOT_CONTAINS_KEY can really just have the same logic - assuming we translate the operators in `convertOperatorToString`, the only difference between these and the default case is the conversion of the LHS (using `convertExpressionToMapKeyColumn` instead of `convertExpressionToString`)

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -250,6 +257,7 @@ private String convertOperatorToString(Operator operator) {
         return "<=";
       case LIKE:
         return REGEX_OPERATOR;
+      case NOT_CONTAINS_KEY:
```
aaron-steinfeld, line 260: These cases aren't right. It seems like they just got parked here so as not to throw since they're not used in the existing code and this is called prematurely. NOT_CONTAINS_KEY becomes `!=` and CONTAINS_KEY becomes `=` (the only difference being that we change the col translation to add on __keys). CONTAINS_KEYVALUE becomes multiple expressions, so could remain until removed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -323,34 +328,31 @@ private String convertExpressionToMapValueColumn(Expression expression) {
...
-        for (int i = 0; i < 2 && i < value.getValue().getStringArrayCount(); i++) {
-          literals[i] =
-              LiteralConstant.newBuilder()
-                  .setValue(
-                      Value.newBuilder().setString(value.getValue().getStringArray(i)).build())
-                  .build();
-        }
+      if (value.getValue().getValueType() == ValueType.STRING_ARRAY
+          && value.getValue().getStringArrayCount() == 2) {
+        literals.set(0, value.getValue().getStringArray(0));
```
aaron-steinfeld, line 337: nit: if we're going to keep it unstructured and use a list, might as well use the same index constants to write this as we use to read it.

aaron-steinfeld (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
sarthak77: Didn't get you on this.

aaron-steinfeld: We read these as via `.get(MAP_KEY_INDEX)` etc., but we write them with plain literals that need to match those constants. Here, we should be using
```java
        literals.set(MAP_KEY_INDEX, value.getValue().getStringArray(MAP_KEY_INDEX));
```
Or better yet, using no constants and just use an object like we do in most java code as discussed earlier.

All relatively moot though as long as this function is slated for deletion in the near term.

sarthak77: ok got it.

Issue #99, opened 2021-11-17
API support for receiving attribute expressions in all places a ColumnIdentifier is currently received


kotharironak:
Based on the [discussion](https://github.com/hypertrace/hypertrace-ui/issues/1099#issuecomment-969376881), accept the new `AttributeExpression` as Query Service Request, and provide similar support for it same as ColumnIdentifer.

```
  AttributeExpression {
    attributeId: String!
    subpath: String
  }
```

Currently, we have `ColumnIdentifer` support in Query Service request in:
- selection
- groupby
- filter
- order by clause

Provide similar support for `AttributeExpression` in QuerySerivce.

E.g of Filter using `ColumnIdentifer`
```
childFilter {
      lhs {
        columnIdentifier {
          columnName: "BACKEND.startTime"
        }
      }
      operator: LT
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631178754903
          }
        }
      }
```

Support for `AttributeExpression`
```
childFilter {
      lhs {
        attributeExpression {
          attributeId: "BACKEND.startTime"
        }
      }
      operator: LT
      rhs {
        literal {
          value {
            valueType: LONG
            long: 1631178754903
          }
        }
      }
```

If you have a test, in both the above cases, it will convert to the same pinot query, and the response builder will have the same response.

## Reference:
See the reference UI ticket - https://github.com/hypertrace/hypertrace-ui/issues/1099 and discussion.

PR #126, opened 2021-12-17
feat : Addition of API support for receiving attribute expressions

sarthak77:
Ref issue: https://github.com/hypertrace/query-service/issues/99

### Files changed : 

- [x] ExecutionContext
- [x] PinotBasedRequestHandler
- [x] ProjectionTransformation
- [x] QueryRequestToPinotSQLConverter

### Testing
- [x] Integration tests ( copy existing ones and check if work for attribute expression as well )
- [x] Integration tests ( new )
- [x] Unit tests


query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryRequestUtil.java
```diff
@@ -139,4 +139,20 @@ public static String getLogicalColumnName(Expression expression) {
                 + " expression type only");
     }
   }
+
+  public static String getAlias(Expression expression) {
```
aaron-steinfeld, line 143: would recommend making this something like getResultName and putting the fallback logic (alias -> column name) in here. In the gateway code I wrote, those two utils - to resolve an attribute ID out of an expression and to resolve a result name out of an expression were used all over the place. It's also helpful to return an optional and let the caller decide whether to throw if missing or not, as you saw in the isMatchingFilter case where you needed to change the logic.

aaron-steinfeld: (from gateway code, as an example)
```java
  public static Optional<String> getSelectionResultName(Expression expression) {
    switch (expression.getValueCase()) {
      case COLUMNIDENTIFIER:
        return Optional.of(
            expression.getColumnIdentifier().getAlias().isEmpty()
                ? expression.getColumnIdentifier().getColumnName()
                : expression.getColumnIdentifier().getAlias());
      case ATTRIBUTE_EXPRESSION:
        return Optional.of(
            expression.getAttributeExpression().getAlias().isEmpty()
                ? expression.getAttributeExpression().getAttributeId()
                : expression.getAttributeExpression().getAlias());
      case FUNCTION:
        FunctionExpression functionExpression = expression.getFunction();
        if (!functionExpression.getAlias().isEmpty()) {
          return Optional.of(functionExpression.getAlias());
        }
        String argumentString =
            functionExpression.getArgumentsList().stream()
                .map(ExpressionReader::getSelectionResultName)
                .flatMap(Optional::stream)
                .collect(Collectors.joining(","));

        return Optional.of(
            String.format("%s_%s", functionExpression.getFunction(), argumentString));
      default:
        return Optional.empty();
    }
  }
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/ExecutionContext.java
```diff
@@ -137,12 +139,12 @@ private ColumnMetadata toColumnMetadata(Expression expression) {
     ValueCase valueCase = expression.getValueCase();
     switch (valueCase) {
       case COLUMNIDENTIFIER:
-        ColumnIdentifier columnIdentifier = expression.getColumnIdentifier();
-        String alias = columnIdentifier.getAlias();
+      case ATTRIBUTE_EXPRESSION:
+        String alias = getAlias(expression);
         if (alias != null && alias.trim().length() > 0) {
```
aaron-steinfeld, line 144: nit: not possible for this code to return null. Below would be semantically equivalent

```suggestion
        if (!alias.isBlank()) {
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryRequestUtil.java
sarthak77: Is optional required since it will always be present ?

aaron-steinfeld: It won't always be present, we don't know what's in the expression. It could be a literal, for example. Optional is preferable to throwing the exception, because it makes it easier for the caller to decide what to do if it's missing - throw a specific exception, ignore etc.

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -96,10 +99,19 @@
 
   private Single<Expression> transformColumnIdentifier(ColumnIdentifier columnIdentifier) {
     return this.projectAttributeIfPossible(columnIdentifier.getColumnName())
-        .map(expression -> this.aliasToMatchOriginal(columnIdentifier, expression))
+        .map(expression -> this.aliasToMatchOriginal(getOriginalKey(columnIdentifier), expression))
         .defaultIfEmpty(Expression.newBuilder().setColumnIdentifier(columnIdentifier).build());
   }
 
+  private Single<Expression> transformAttributeExpression(AttributeExpression attributeExpression) {
+    return this.projectAttributeIfPossible(attributeExpression.getAttributeId())
```
aaron-steinfeld, line 107: This would drop the subpath, need to add it back on

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -417,4 +432,14 @@ private Filter updateFilterForComplexAttributeExpressionFromFilter(Filter origin
         .map(QueryRequestUtil::createContainsKeyFilter)
         .collect(Collectors.toList());
   }
+
+  private String getOriginalKey(AttributeExpression attributeExpression) {
+    String alias = attributeExpression.getAlias();
```
aaron-steinfeld, line 437: We have a function for this now I believe, can share it right?

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryRequestUtil.java
sarthak77: got it.

sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -225,10 +226,12 @@ private long getRequestStartTime(Filter filter) {
   }
 
   private boolean lhsIsStartTimeAttribute(Expression lhs) {
-    return lhs.hasColumnIdentifier()
-        && startTimeAttributeName
-            .map(attributeName -> attributeName.equals(lhs.getColumnIdentifier().getColumnName()))
-            .orElse(false);
+    if (lhs.hasColumnIdentifier() || lhs.hasAttributeExpression()) {
```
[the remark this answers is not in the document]
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -274,14 +277,15 @@ private boolean rhsHasLongValue(Expression rhs) {
    */
   private boolean doesSingleViewFilterMatchLeafQueryFilter(
       Map<String, ViewColumnFilter> viewFilterMap, Filter queryFilter) {
-    if (queryFilter.getLhs().getValueCase() != ValueCase.COLUMNIDENTIFIER) {
+    if (queryFilter.getLhs().getValueCase() != ValueCase.COLUMNIDENTIFIER
```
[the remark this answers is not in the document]
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/ExecutionContext.java
```diff
@@ -219,12 +225,12 @@ private void extractColumns(List<String> columns, Expression expression) {
...
 
-    // if not leaf filter, then return false
-    if (!filter.hasLhs()) {
+    Optional<String> logicalColumnName = getLogicalColumnName(filter.getLhs());
+    if (logicalColumnName.isEmpty()) {
       return false;
     }
 
-    return column.equals(getLogicalColumnName(filter.getLhs()))
+    return column.equals(logicalColumnName.get())
```
aaron-steinfeld, line 233: In general, `Optional.get()` is an antipattern, it shouldn't even be part of the API in my opinion. If you want to work imperatively, you should use either `orElseThrow()` (which is identical in behavior to `.get()`, but conveys the behavior better, since `get()` will throw too), or `orElse[Get]()` to provide a default. Other options include functional style:
```java
return getLogicalColumnName(filter.getLhs())
  .map(column::equals)
  .orElse(false) && ...
```
or comparing optionals directly rather than unwrapping to do the comparison:
```java
return Optional.of(column).equals(getLogicalColumnName(filter.getLhs())) && ...
```

The benefit of both of these examples is they prevent logic separation. In the current code, line 233 relies on an earlier check (line 229) to behave correctly. This is a fairly trivial case, but anywhere we can reduce logic separation, we're reducing the chance of bugs when a later maintainer comes in and changes one place without realizing another relies on it.

aaron-steinfeld: This is a nit in this location btw, but trying to give more context why the optionals are actually more maintainable than throwing directly in `getLogicalColumnName`.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -277,16 +273,17 @@ private boolean rhsHasLongValue(Expression rhs) {
...
       Map<String, ViewColumnFilter> viewFilterMap, Filter queryFilter) {
-    if (queryFilter.getLhs().getValueCase() != ValueCase.COLUMNIDENTIFIER
-        && queryFilter.getLhs().getValueCase() != ValueCase.ATTRIBUTE_EXPRESSION) {
+
+    if (queryFilter.getOperator() != Operator.IN && queryFilter.getOperator() != Operator.EQ) {
       return false;
     }
-    if (queryFilter.getOperator() != Operator.IN && queryFilter.getOperator() != Operator.EQ) {
+
+    Optional<String> columnName = getLogicalColumnName(queryFilter.getLhs());
```
aaron-steinfeld, line 281: nit: could be combined with below check, e.g.
```java
ViewColumnFilter viewColumnFilter = columnName.map(viewFilterMap::get).orElse(null);
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -211,7 +211,7 @@ private Expression handleValueConversionForLiteralExpression(Expression lhs, Exp
       return rhs;
     }
 
-    String lhsColumnName = getLogicalColumnName(lhs);
+    String lhsColumnName = getLogicalColumnName(lhs).orElseThrow(IllegalArgumentException::new);
```
aaron-steinfeld, line 214: nit: `orElseThrow()` is probably sufficient throughout, no need to convert them all to `IllegalArgumentException` - without a message it doesn't really provide any extra information.

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestEligibilityValidator.java
```diff
@@ -89,12 +90,14 @@ private boolean selectionAndGroupByOnDifferentColumn(
     Set<String> selections =
         selectionList.stream()
             .map(QueryRequestUtil::getLogicalColumnName)
+            .map(Optional::get)
```
aaron-steinfeld, line 93: If you want it to error on unexpected
```suggestion
            .map(Optional::orElseThrow)
```
If you want it to drop and recover on unexpected
```suggestion
            .flatMap(Optional::stream)
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
aaron-steinfeld: This bug is the blocker btw - the rest are code style nits.

query-service-impl/src/main/java/org/hypertrace/core/query/service/ExecutionContext.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/prometheus/QueryRequestEligibilityValidator.java
sarthak77: resolved

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -274,15 +273,13 @@ private boolean rhsHasLongValue(Expression rhs) {
...
-    }
+
     if (queryFilter.getOperator() != Operator.IN && queryFilter.getOperator() != Operator.EQ) {
       return false;
     }
 
-    String columnName = queryFilter.getLhs().getColumnIdentifier().getColumnName();
-    ViewColumnFilter viewColumnFilter = viewFilterMap.get(columnName);
+    ViewColumnFilter viewColumnFilter =
+        viewFilterMap.get(getLogicalColumnName(queryFilter.getLhs()).orElse(null));
```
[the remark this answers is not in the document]
aaron-steinfeld: Should work in most map impls, but I'm still partial to the original suggestion:
https://github.com/hypertrace/query-service/pull/126#discussion_r773192690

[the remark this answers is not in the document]
aaron-steinfeld: > Throws:
ClassCastException – if the key is of an inappropriate type for this map (optional)
NullPointerException – if the specified key is null and this map does not permit null keys (optional)

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
sarthak77: That accepts expression.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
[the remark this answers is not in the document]
sarthak77: This is defined in `ViewDefinition.java`

```
    // Check if there are any view filters. If there are multiple filters, they all will
    // be AND'ed together.
    final Map<String, ViewColumnFilter> filterMap = new HashMap<>();
    if (config.hasPath(FILTERS_CONFIG_KEY)) {
      for (Config filterConfig : config.getConfigList(FILTERS_CONFIG_KEY)) {
        filterMap.put(
            filterConfig.getString(COLUMN_CONFIG_KEY), ViewColumnFilter.from(filterConfig));
      }
    }
```

kotharironak:
@sarthak77 Can you add the integration test as discussed in the morning?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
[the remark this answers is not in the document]
kotharironak: > Should work in most map impls, but I'm still partial to the original suggestion: [#126 (comment)](https://github.com/hypertrace/query-service/pull/126#discussion_r773192690)

Yes, that would be better.

```
ViewColumnFilter viewColumnFilter =
        getLogicalColumnName(queryFilter.getLhs()).map(viewFilterMap::get).orElse(null);
```

query-service/src/integrationTest/java/org/hypertrace/core/query/service/QueryServiceTestUtils.java
```diff
@@ -103,10 +113,38 @@ public static Expression createStringLiteralValueExpression(String value) {
...
-            AttributeExpression.newBuilder().setAttributeId(attributeId).setSubpath(subPath));
+  public static QueryRequest getAttributeExpressionQuery(QueryRequest originalQueryRequest)
+      throws InvalidProtocolBufferException {
+    // Serialize into json.
+    String json =
+        JsonFormat.printer().omittingInsignificantWhitespace().print(originalQueryRequest);
+    System.out.println(json);
+
+    // Change for attribute Expression
+    json = json.replaceAll("columnIdentifier", "attributeExpression");
```
kotharironak, line 124: We don't have any existing columnIdentifier queries with map field with CONTAINS_KEY_VALUE, right? If, so, it will fail the transformation.

query-service/src/integrationTest/resources/attribute-expression-test-queries/query2.json
```diff
@@ -0,0 +1,85 @@
...
+      "functionName": "AVG",
+      "arguments": [{
+        "attributeExpression": {
+          "attributeId": "EVENT.duration"
+        }
+      }],
+      "alias": "avg_duration"
+    }
+  }, {
+    "attributeExpression": {
```
kotharironak, line 74: Can we also add aggregation function on map attribute expression with subpath - e.g `sum(spanTags__Keys, 'span.kind', spanTags__values) ?

query-service/src/integrationTest/resources/attribute-expression-test-queries/query2.json
```diff
@@ -0,0 +1,85 @@
+{
```
kotharironak, line 1: Can we also add one more query say `query3.json` with selection having map attribute expression w/o subpath?

kotharironak: We should also add test for `NOT_CONTAINS_KEY`.

query-service/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ dependencies {
   implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.6.2")
   implementation("org.hypertrace.core.serviceframework:platform-service-framework:0.1.28")
   implementation("org.slf4j:slf4j-api:1.7.30")
+  implementation("com.google.protobuf:protobuf-java-util:3.17.3")
```
aaron-steinfeld, line 16: you need this in your int tests, not at runtime I'm assuming

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
aaron-steinfeld: Fixed in https://github.com/hypertrace/query-service/pull/129

query-service/src/integrationTest/resources/attribute-expression-test-queries/query2.json
aaron-steinfeld: Please add these tests in followup

query-service/src/integrationTest/resources/attribute-expression-test-queries/query2.json
aaron-steinfeld: Please add these tests in followup

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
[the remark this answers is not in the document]
aaron-steinfeld: Fixed in https://github.com/hypertrace/query-service/pull/129

aaron-steinfeld (approved):

query-service/src/integrationTest/resources/attribute-expression-test-queries/query2.json
sarthak77: Does this sample query work for this case ?
```
select sum(mapValue(tags__KEYS, 'otel.status_code', tags__VALUES)) from spanEventView
```

sarthak77: added

query-service/src/integrationTest/resources/attribute-expression-test-queries/query2.json
sarthak77: added

sarthak77: added

PR #129, opened 2021-12-23
fix: support for attribute expressions in selections

aaron-steinfeld:
## Description

- Add support for attribute expressions in selections, aggregations.
- Fix bug where subpath is used in projection.
- Refactor separate distinct query transformations.

### Testing

- Updated unit tests
- Ran existing Integration tests


aaron-steinfeld (commented):
Looks bigger than it is. Basically just refactored/split the transformations and their tests, fixed a few small bugs that had been pending from prior reviews.

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionSubpathExistsFilteringTransformation.java
```diff
@@ -0,0 +1,134 @@
...
+    return super.rebuildRequest(
+        original, selections, aggregations, updatedFilter, groupBys, orderBys);
+  }
+
+  /*
+   * We need the CONTAINS_KEY filter in all filters and order bys dealing with complex
+   * attribute expressions as Pinot gives error if particular key is absent. Rest all work fine.
+   * To handle order bys, we add the corresponding filter at the top and 'AND' it with the main filter.
+   * To handle filter, we modify each filter (say filter1) as : "CONTAINS_KEY AND filter1".
+   */
```
sarthak77, line 54:
```suggestion
  /*
   * We need the CONTAINS_KEY filter in all filters, selections, and order bys dealing with complex
   * attribute expressions as Pinot gives an error if a particular key is absent. Rest all work fine.
   * To handle order bys and selections, we add the corresponding filter at the top and 'AND' it with the main filter.
   * To handle filter, we modify each filter (say filter1) as: "CONTAINS_KEY AND filter1".
   */
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionSubpathExistsFilteringTransformation.java
```diff
@@ -0,0 +1,134 @@
...
+        .map(QueryRequestUtil::createContainsKeyFilter)
+        .collect(Collectors.toList());
+  }
+
+  private Stream<AttributeExpression> getAnyAttributeExpression(Expression expression) {
+    if (expression.hasFunction()) {
+      return expression.getFunction().getArgumentsList().stream()
+          .flatMap(this::getAnyAttributeExpression);
+    }
+    return Stream.of(expression)
```
sarthak77, line 130: nit : Can we have this in else block ?

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionSubpathExistsFilteringTransformationTest.java
```diff
@@ -0,0 +1,235 @@
...
+            .addOrderBy(createOrderByExpression(spanTag1, SortOrder.ASC))
+            .addOrderBy(createOrderByExpression(spanTag2, SortOrder.ASC))
+            .build();
+
+    QueryRequest expectedTransform =
+        QueryRequest.newBuilder()
+            .setFilter(
+                createCompositeFilter(
+                    Operator.AND,
+                    createContainsKeyFilter("Span.tags", "span.kind"),
```
sarthak77, line 225: Since w are using a set now, is the order of filters ensured? Like sometimes this test may pass and sometimes it may fail? Similarly for other tests.

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
```diff
@@ -0,0 +1,41 @@
...
+  protected Single<Expression> transformColumnIdentifier(ColumnIdentifier columnIdentifier) {
+    Expression.Builder expressionBuilder =
+        QueryRequestUtil.createSimpleAttributeExpression(columnIdentifier.getColumnName());
+    if (columnIdentifier.getAlias().isBlank()) {
+      return Single.just(expressionBuilder.build());
+    }
+
+    return Single.just(
+        expressionBuilder
+            .setAttributeExpression(
```
sarthak77, line 35: Don't we need to unset the present attribute expression first ?

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformationTest.java
```diff
@@ -0,0 +1,75 @@
...
+@ExtendWith(MockitoExtension.class)
+class AttributeExpressionNormalizationTransformationTest {
+
+  @Mock QueryTransformationContext mockTransformationContext;
+
+  private final AttributeExpressionNormalizationTransformation transformation =
+      new AttributeExpressionNormalizationTransformation();
+
+  @Test
+  void columnInAllPositions() {
```
sarthak77, line 31: nit : transformColumnInAllPositions ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionSubpathExistsFilteringTransformation.java
aaron-steinfeld: style preference - I always unwrap else blocks if the if statement returns to reduce nesting. Here's a lint rule (for a different language, but same pattern) that gives a tad more context: https://palantir.github.io/tslint/rules/unnecessary-else/

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformationTest.java
aaron-steinfeld:
```suggestion
  void transformColumnInAllPositions() {
```

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionSubpathExistsFilteringTransformationTest.java
aaron-steinfeld: It's an ordered set, there's no non-determinism (if you see any, please flag it - ImmutableSet or LinkedHashSet at least both maintain order)..

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
aaron-steinfeld: Not sure I follow. There is no attribute expression to start, we've created one inside an expression on line 27. Here, we update that attribute expression to include an alias.

sarthak77: Yes, one created on line 27. Don't we need to unset that ?

aaron-steinfeld: We're overwriting it - an expression can only contain one attribute expression, so we're reading it's value out, updating it and writing it back over the original.

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Override
+  protected Logger getLogger() {
+    return log;
+  }
+
+  @Override
+  protected Single<Expression> transformColumnIdentifier(ColumnIdentifier columnIdentifier) {
+    Expression.Builder expressionBuilder =
+        QueryRequestUtil.createSimpleAttributeExpression(columnIdentifier.getColumnName());
```
kotharironak, line 28: What will happen with column identifier with `CONTAIN_KEY_VALUE` operator? So, after this change, would that be broken or taken care of in order deployment?

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformationTest.java
```diff
@@ -0,0 +1,75 @@
...
+            .addSelection(createAliasedColumnExpression("select-col", "select-alias"))
+            .addAggregation(
+                createAliasedFunctionExpression(
+                    QueryFunctionConstants.QUERY_FUNCTION_SUM,
+                    "agg-alias",
+                    createColumnExpression("agg-col").build()))
+            .setFilter(
+                Filter.newBuilder()
+                    .setLhs(createAliasedColumnExpression("filter-col", "filter-alias"))
+                    .setOperator(Operator.EQ)
```
kotharironak, line 43: This looks fine to me, but it seems to me that we might break `CONTAIN_KEY_VALUE` in transition.

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
```diff
@@ -57,96 +47,33 @@
...
-    return this.transformExpression(orderBy.getExpression())
-        .map(orderBy.toBuilder()::setExpression)
-        .map(OrderByExpression.Builder::build);
-  }
-
-  private Single<Filter> transformFilter(Filter filter) {
-    if (filter.equals(Filter.getDefaultInstance())) {
-      return Single.just(filter);
+  private Expression addSubpathOrThrow(Expression projectedExpression, String subpath) {
+    if (!QueryRequestUtil.isSimpleAttributeExpression(projectedExpression)) {
```
kotharironak, line 69: Can you give one example or add a test for this when we hit this condition?

A projected attribute is already having subpath?

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
aaron-steinfeld: I'd expect it to translate same as anything else - something like:
```json
{
      "lhs": {
        "attributeExpression": {
          "attributeId": "EVENT.spanTags"
        }
      },
      "operator": "CONTAINS_KEY_VALUE",
      "rhs": {
        "literal": {
          "value": {
            "stringArray": ["span.kind", "client"]
          }
        }
      }
    }
```

If that doesn't work (we can and should add tests for it), then we'd be broken with or without this change, because we don't expect column identifiers to be coming at all once the gateway change is merged.

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformationTest.java
aaron-steinfeld: You mentioned this in the previous comment, but I think I may be missing how this would break. Could you be more specific?

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
aaron-steinfeld: To elaborate on my current understanding, I thought the support we didn't add WRT `CONTAIN_KEY_VALUE ` was for it to take an attribute expression with subpath on the LHS and a single value on the RHS. Likewise, we didn't add support for `[NOT_]CONTAINS_KEY` with a subpath'd expression and no args. Given that, any arriving expression using a column identifier necessarily has to use the old RHS format so should translate identically.

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
aaron-steinfeld: Yeah, will add a test. As a quick example, imagine if we rename a column and so use a simple projection to point one attribute name at another. In this hypothetical, an old name of `EVENT.spanTags` becomes `EVENT.attributes`. An arriving attribute expression may still be using the old name, so when we project
```json
       "attributeExpression": {
          "attributeId": "EVENT.spanTags",
          "subpath": "span.kind"
        }
```

the projection is only based on the attribute itself, so above we first project:
```json
       "attributeExpression": {
          "attributeId": "EVENT.attributes"
        }
```
then this function is meant to restore the subpath part giving us our final projection:
```json
       "attributeExpression": {
          "attributeId": "EVENT.attributes",
          "subpath": "span.kind"
        }
```

It currently throws if the projection did not result in another simple attribute expression, since if the result was already a complex expression (either a function or map selection), it would be more difficult/potentially impossible to support (to be honest, I didn't go too far down that path - we don't have that use case at the moment and can cross that bridge when/if we get there).

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
sarthak77: > I'd expect it to translate same as anything else - something like:

Currently unit test is not there for such case as it was being deprecated. Can add it in my PR.

sarthak77: Added in this PR.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/MigrationTest.java
```diff
@@ -226,6 +228,35 @@ public void testQueryWithNotContainsKeyOperator() {
         executionContext);
   }
 
+  @Test
```
sarthak77, line 231: CC @kotharironak @aaron-steinfeld

query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java
```diff
@@ -96,8 +96,7 @@ public static void setup() throws Exception {
         new GenericContainer<>(DockerImageName.parse("hypertrace/pinot-servicemanager:main"))
             .withNetwork(network)
             .withNetworkAliases("pinot-controller", "pinot-server", "pinot-broker")
-            .withExposedPorts(8099)
-            .withExposedPorts(9000)
+            .withExposedPorts(8099, 9000)
```
aaron-steinfeld, line 99: This was a pre-existing bug with test config, the second one overwrites the first, it takes varargs instead. It was previously working due to a bug in test containers that was fixed when we upgraded to fix vulnerable deps.

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryTransformationPipeline.java
```diff
@@ -24,6 +24,7 @@
     QueryTransformationContext transformationContext =
         new DefaultQueryTransformationContext(tenantId);
     return Observable.fromIterable(transformations)
+        .sorted()
```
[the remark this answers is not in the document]
aaron-steinfeld: As part of this PR, we put in the concept of ordered transformations, so this ensures their order (since mutli injection does not).

query-service-impl/src/main/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformation.java
kotharironak: Yes, For this - https://github.com/hypertrace/query-service/pull/129/files#r775325504,  I re-looked the code, and it is taken care of here in method `convertExpressionToMapLiterals` for both `string_array` and for `string`.

So, it looks fine to me.

query-service-impl/src/test/java/org/hypertrace/core/query/service/attribubteexpression/AttributeExpressionNormalizationTransformationTest.java
kotharironak: I was thinking that `string_array` part is broken as you described here - https://github.com/hypertrace/query-service/pull/129#discussion_r775325504 - if we translate as is. But, I relooked, and the method `convertExpressionToMapLiterals` in `QueryRequestToPinotSQLConverter` takes care of it for now.

kotharironak (approved):

kotharironak:
@aaron-steinfeld let me know if we are upgrading to log4j version? I will re-approve.

aaron-steinfeld:
Will upgrade it in a following PR

query-service-impl/src/main/java/org/hypertrace/core/query/service/projection/ProjectionTransformation.java
aaron-steinfeld: Test added in #132

aaron-steinfeld:
> Will upgrade it in a following PR

https://github.com/hypertrace/query-service/pull/132

Issue #130, opened 2021-12-24
Add integ tests


sarthak77:
Add integration tests to test the new changes for attribute expression. 

- [x] https://github.com/hypertrace/query-service/pull/126#discussion_r774504121
- [x] https://github.com/hypertrace/query-service/pull/126#discussion_r774504647

PR #131, opened 2021-12-24
Add integ tests

sarthak77:
Ref issue : https://github.com/hypertrace/query-service/issues/130


query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java
```diff
@@ -447,6 +447,11 @@ public void testAttributeExpressionQueries(
       throws IOException {
     return Stream.of(
         Arguments.arguments(buildQueryFromJsonFile("query1.json"), 10, "server"),
-        Arguments.arguments(buildQueryFromJsonFile("query2.json"), 2, "server"));
+        Arguments.arguments(buildQueryFromJsonFile("query2.json"), 2, "server"),
+        Arguments.arguments(buildQueryFromJsonFile("query3.json"), 1, "0.0"),
+        Arguments.arguments(
+            buildQueryFromJsonFile("query4.json"),
+            10,
+            "{\"internal.span.format\":\"jaeger\",\"http.status_code\":\"200\",\"component\":\"net/http\",\"span.kind\":\"server\",\"sampler.type\":\"const\",\"sampler.param\":\"1\",\"http.url\":\"/dispatch?customer=392&nonse=0.2664909002836733\",\"servicename\":\"frontend\",\"http.method\":\"GET\"}"));
```
kotharironak, line 455: Can you also put that in json file as `query4-expected-response.json`? (its not redable)

sarthak77: It is coming in this format only. It is a string.

sarthak77: created a new test for this case.

query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java
```diff
@@ -419,6 +419,19 @@ public void testAttributeExpressionQueries(
...
+  public void testNotContainsKeyQuery() throws IOException {
+    LOG.info("Not contains key query");
+    Iterator<ResultSetChunk> itr =
+        queryServiceClient.executeQuery(
+            buildQueryFromJsonFile("query4.json"), TENANT_ID_MAP, 10000);
+    List<ResultSetChunk> list = Streams.stream(itr).collect(Collectors.toList());
+    List<Row> rows = list.get(0).getRowList();
+    assertEquals(10, rows.size());
+    assertEquals(
+        "\"http.status_code\":\"200\"", rows.get(0).getColumn(0).getString().split(",")[1]);
```
kotharironak, line 432: Are all the rows are having the same values? Would this always be true that the first-row first column will have this value?

sarthak77: >Are all the rows are having the same values?

No

>Would this always be true that the first-row first column will have this value?

Should be as we are generating same load ?

query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java
```diff
@@ -419,6 +419,19 @@ public void testAttributeExpressionQueries(
     assertEquals(expectedValue, rows.get(0).getColumn(0).getString());
   }
 
+  @Test
+  public void testNotContainsKeyQuery() throws IOException {
+    LOG.info("Not contains key query");
+    Iterator<ResultSetChunk> itr =
+        queryServiceClient.executeQuery(
+            buildQueryFromJsonFile("query4.json"), TENANT_ID_MAP, 10000);
```
rish691, line 427: is there a 1-1 mapping between the test method and query file, if yes then maybe you can name the query file after the method name?

rish691 (approved):

PR #135, opened 2022-03-16
feat: add configurable limit validation

aaron-steinfeld:
## Description
This PR adds configurable limit validation. It supports setting a minimum limit (preventing things like negative values, or unlimited queries), maximum limit, and a mode. The three modes are WARN (which logs limit breaches, but does not reject them), ERROR (which... errors) and DISABLED (probably self explanatory). Min and Max values are inclusive. Defaults are set to a limit of 1-10K, in warn mode. In practice this means warning logs will fire for any unlimited query, or a query with a limit > 10K.

### Testing
Added new unit tests for all new code, as well as updating existing test paths.

### Checklist:

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/validation/LimitValidation.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/validation/LimitValidation.java
@@ -1,0 +1,60 @@
+package org.hypertrace.core.query.service.validation;
+
+import io.grpc.Status;
+import io.reactivex.rxjava3.core.Completable;
+import javax.inject.Inject;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.core.query.service.QueryServiceConfig;
+import org.hypertrace.core.query.service.QueryServiceConfig.LimitValidationConfig;
+import org.hypertrace.core.query.service.api.QueryRequest;
+
+@Slf4j
+class LimitValidation implements QueryValidation {
+
+  LimitValidationConfig config;
+
+  @Inject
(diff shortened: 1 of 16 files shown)


aaron-steinfeld:
Put snyk fixes into separate PR: https://github.com/hypertrace/query-service/pull/136

query-service-impl/src/main/java/org/hypertrace/core/query/service/validation/LimitValidation.java
```diff
@@ -0,0 +1,60 @@
...
+      case WARN:
+        if (isInvalidLimit(queryRequest.getLimit())) {
+          log.warn(
+              generateErrorMessageForLimit(queryRequest.getLimit())
+                  + ". Allowing due to warn mode.",
+              queryRequest.getLimit(),
+              config.getMin(),
+              config.getMax());
+        }
+        return Completable.complete();
```
kotharironak, line 44: Thought: was just thinking in the warning mode for invalid limit request, should we reset the boundary to configured min/max limit?

kotharironak (approved):
lgtm

avinashkolluru (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/validation/LimitValidation.java
aaron-steinfeld: > Thought: was just thinking in the warning mode for invalid limit request, should we reset the boundary to configured min/max limit?

I'd probably treat that as a separate mode (or even the behavior in error mode). To start, I'd like to find the places in the platform/UI firing queries with no limits or large limits to make sure no logic changes are needed there. For example, a UI page showing a list of events without limits shouldn't just be limited, it should be paged on the UI side.

PR #141, opened 2022-05-12
DISTINCTCOUNTMV

japoorv:
Adding support for `DISTINCTCOUNTMV` for multi-value columns (arrays)


query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
```diff
@@ -0,0 +1,79 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.attribute.service.cachingclient.CachingAttributeClient;
+import org.hypertrace.core.attribute.service.v1.AttributeKind;
+import org.hypertrace.core.query.service.AbstractQueryTransformation;
+import org.hypertrace.core.query.service.api.Expression;
+import org.hypertrace.core.query.service.api.Function;
+import org.slf4j.Logger;
+
+@Slf4j
+public class DistinctCountFunctionTransformation extends AbstractQueryTransformation {
```
skjindal93, line 21: How about we make this class more generic for transforming any multi valued attribute?

japoorv: I see, so something like `MULTIVALUE` transformation class which takes care of all these kind of transformations ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
```diff
@@ -0,0 +1,79 @@
...
+                    ? functionBuilder.setFunctionName(DISTINCT_COUNT_MV)
+                    : functionBuilder);
+  }
+
+  private boolean isArray(AttributeKind attributeKind) {
+    return List.of(TYPE_STRING_ARRAY, TYPE_BOOL_ARRAY, TYPE_DOUBLE_ARRAY, TYPE_INT64_ARRAY)
+        .contains(attributeKind);
+  }
+
+  private Single<String> getFirstAttributeId(Function.Builder functionBuilder) {
```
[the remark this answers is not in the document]
japoorv: The attribute which is defined in the first argument in the arguments list of the function. 
(assumption being that the transformation would happen only on distinct count of a singleton attribute expression list), something like 
`function {
name: DISTINCTCOUNT
arguments: [AttributeExpression(....)]
} `

query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
```diff
@@ -0,0 +1,79 @@
...
+    return Observable.fromIterable(expressionList)
+        .concatMapSingle(this::transformExpression)
+        .toList();
+  }
+
+  // Checking if the function is of the form DISTINCT_COUNT(ATTRIBUTE_EXPRESSION)
+  private boolean isTransformable(Function.Builder builder) {
+    return builder.getFunctionName().equals(DISTINCT_COUNT)
+        && builder.getArgumentsList().size() == 1
+        && builder.getArgumentsList().get(0).hasAttributeExpression();
```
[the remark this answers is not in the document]
japoorv: Don't really know how column identifier is used. 
I see that it has a `name` field, so does that mean it takes the column name ? 
If yes how does that differ from `AttributeExpression`

query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
japoorv: refactored

avinashkolluru:
Looks good to me

avinashkolluru (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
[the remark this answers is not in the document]
skjindal93: Yes, the name corresponds to the column name. `AttributeExpression` is a newer way of `ColumnIdentifier` but, with additional support for map fields

`ColumnIdentifier` should ideally be deprecated, but we haven't moved it out

query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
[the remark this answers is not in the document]
skjindal93: Fair assumption

query-service-impl/src/main/java/org/hypertrace/core/query/service/function/DistinctCountFunctionTransformation.java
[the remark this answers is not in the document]
avinashkolluru: Yes. I think we can.

query-service-impl/src/main/java/org/hypertrace/core/query/service/multivalue/MultiValueTransformation.java
```diff
@@ -0,0 +1,41 @@
...
+  MultiValueTransformation(FunctionTransformation functionTransformation) {
+    this.functionTransformation = functionTransformation;
+  }
+
+  @Override
+  protected Logger getLogger() {
+    return log;
+  }
+
+  @Override
```
aaron-steinfeld, line 27: we're copy pasting the logic of the parent class in this function and the one below - instead, transform the function only and then send it back to super.

query-service-impl/src/main/java/org/hypertrace/core/query/service/multivalue/MultiValueTransformation.java
```diff
@@ -0,0 +1,41 @@
...
+  protected Single<Expression> transformFunction(Function function) {
+    return this.transformExpressionList(function.getArgumentsList())
+        .map(function.toBuilder().clearArguments()::addAllArguments)
+        .flatMap(functionTransformation::transformFunction)
```
aaron-steinfeld, line 31: This class is a pinot specific transformation, so does not belong as a query transformation, it belongs in the pinot converter. This is still at the abstract, data store agnostic level.

query-service-impl/src/main/java/org/hypertrace/core/query/service/multivalue/FunctionTransformation.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final List<AttributeKind> ARRAY_KINDS =
+      List.of(TYPE_STRING_ARRAY, TYPE_BOOL_ARRAY, TYPE_DOUBLE_ARRAY, TYPE_INT64_ARRAY);
+  private final CachingAttributeClient attributeClient;
+
+  @Inject
+  public FunctionTransformation(CachingAttributeClient attributeClient) {
+    this.attributeClient = attributeClient;
+  }
+
+  public Single<Function.Builder> transformFunction(Function.Builder builder) {
```
aaron-steinfeld, line 27: should not accept and mutate and return a builder - accept a function, return a function.

query-service-impl/src/main/java/org/hypertrace/core/query/service/multivalue/FunctionTransformation.java
```diff
@@ -0,0 +1,58 @@
...
+    switch (builder.getFunctionName()) {
+      case DISTINCT_COUNT:
+        // Checking if the function is of the form DISTINCT_COUNT(ATTRIBUTE_EXPRESSION)
+        if (builder.getArgumentsList().size() == 1
+            && builder.getArgumentsList().get(0).hasAttributeExpression()) {
```
aaron-steinfeld, line 32: most of these little bits of logic already exist in `QueryRequestUtil` - `isSimpleAttributeExpression` (accounts for column identifier case too) and `getLogicalColumnName` gets the ID for either.

query-service-impl/src/main/java/org/hypertrace/core/query/service/multivalue/MultiValueTransformation.java
japoorv: Understood.
Just an implementation question here, since we are also checking the `attribute-service client` for attribute info we need a reactive way of converting in pinot converters. 
What I see right now is that since all the converters are mostly converting to PQL/SQL everything is straightforward. However if we introduce this we need a `get` on `caching client` which would return a single, so either we do a blocking get on that or we convert the entire implementation (fetching result from Pinot in the PinotRequestHandler) to a reactive impl.

aaron-steinfeld: I would look at putting the logic in `PinotFunctionConverter`, which is where we go from generic function names to the pinot specific.

> if we introduce this we need a get on caching client which would return a single, so either we do a blocking get on that or we convert the entire implementation

True. I wouldn't convert the entire implementation. I'd first look to see if we can actually get this data synchronously. For example, ViewDefinition.getColumnType may give you that same information. If not, blockingGet would be acceptable here because we're already doing other blocking remote calls anyway. We don't want to do a blocking get when we're working in async code, but this is all sync code anyway.

japoorv: Explored this, seems fine except one issue. 
The `PinotFunctionConverter` and most of the request handler classes are being initialised manually (rather than guice injection) due to which we would have to change the definition of all of them (along with all the test classes) to allow the attribute client in the `PinotFunctionConverter` to be passed as a constructor argument. Is there some way to avoid this ?

japoorv: So just checked the ViewDefinition code, it is constructing the spec based on the explicit mention of whether the field is MAP/bytes (https://github.com/hypertrace/query-service/blob/1730c1ebfc6ade862f25c042a0bd2189960a99f6/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewDefinition.java#L111). 
In this case when it is multi value, one option is to implement the Attribute Client call in the view definition (https://github.com/hypertrace/query-service/blob/1730c1ebfc6ade862f25c042a0bd2189960a99f6/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewDefinition.java#L110) which again brings to our original problem of attribute client. The other option is to add an extra config in the handler for recognising multi valued columns.

aaron-steinfeld: > one option is to implement the Attribute Client call in the view definition

Wouldn't do that - separate concerns (attribute vs pinot definitions).

Sounds like the option is either to thread the attribute client through, or the last thing you mentioned, add more config (would avoid this, since it's duplicate config and easy to get out of sync).  Another option could be to add that info when building the execution context - the attribute client might be available at that point and I believe we put other metadata in there.

PR #143, opened 2022-06-23
feat: support configuring distinct count agg function

aaron-steinfeld:
## Description
- Adds the ability to configure the aggregation used for DISTINCTCOUNT, defaulting to the existing implementation which is the exact computation. Behavior of various versions of this function are described in https://docs.pinot.apache.org/users/user-guide-query/how-to-handle-unique-counting
- Updates the avg rate function to use the `/` operator rather than the existing `DIV()`. This is better optimized by Pinot.
- Refactored config of aggregation configuration for easier extendability

### Testing
Ran and updated unit tests

### Checklist:

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
@@ -1,0 +1,38 @@
+package org.hypertrace.core.query.service.pinot.converters;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@Value
+@AllArgsConstructor
+public class PinotFunctionConverterConfig {
+
+  private static final String PERCENTILE_AGGREGATION_FUNCTION_CONFIG = "percentileAggFunction";
+  private static final String DISTINCT_COUNT_AGGREGATION_FUNCTION_CONFIG =
+      "distinctCountAggFunction";
+  private static final String DEFAULT_PERCENTILE_AGGREGATION_FUNCTION = "PERCENTILETDIGEST";
+  private static final String DEFAULT_DISTINCT_COUNT_AGGREGATION_FUNCTION = "DISTINCTCOUNT";
+
+  String percentileAggregationFunction;
+  String distinctCountAggregationFunction;
+
+  public PinotFunctionConverterConfig(Config config) {
+    if (config.hasPath(PERCENTILE_AGGREGATION_FUNCTION_CONFIG)) {
+      this.percentileAggregationFunction = config.getString(PERCENTILE_AGGREGATION_FUNCTION_CONFIG);
+    } else {
+      this.percentileAggregationFunction = DEFAULT_PERCENTILE_AGGREGATION_FUNCTION;
+    }
+    if (config.hasPath(DISTINCT_COUNT_AGGREGATION_FUNCTION_CONFIG)) {
+      this.distinctCountAggregationFunction =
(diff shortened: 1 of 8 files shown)


aaron-steinfeld:
Verified this is broken on main. This is why reproducible tests are good :( https://github.com/hypertrace/query-service/runs/7032109114?check_suite_focus=true

Guessing the integration tests are not set up to support the current version of pinot, will see if I can poke some things.

aaron-steinfeld:
Confirmed, introduced by https://github.com/hypertrace/pinot/releases/tag/0.5.0 . Pinning to the previous tag completes successfully.

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterTest.java
```diff
@@ -211,7 +211,7 @@ void convertAvgRateFunction() {
         .thenReturn(Optional.of(Duration.ofSeconds(10)));
 
     assertEquals(
-        "SUM(DIV(foo, 2.0))",
+        "SUM(foo / 2.0)",
```
kotharironak, line 214: Just curious: Do we have any value level test that compares prior (SUM(DIV(foo, 2.0)) vs SUM(foo / 2.0, resulting in the same values?

kotharironak (commented):
overall lgtm

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterTest.java
aaron-steinfeld: That would be testing pinot's functionality - that `DIV` and `/` both actually... divide. Not sure it's worth writing a test, but we can verify by hand real quick.

aaron-steinfeld:
```
select sum(num_calls), sum(num_calls / 60.0), sum(div(num_calls, 60.0)) ...
```

image

kotharironak:
> Confirmed, introduced by https://github.com/hypertrace/pinot/releases/tag/0.5.0 . Pinning to the previous tag completes successfully.

So, are you pinning to the previous tag `hypertrace/pinot-servicemanager:main` to fix this test now?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Value
+@AllArgsConstructor
+public class PinotFunctionConverterConfig {
+
+  private static final String PERCENTILE_AGGREGATION_FUNCTION_CONFIG = "percentileAggFunction";
+  private static final String DISTINCT_COUNT_AGGREGATION_FUNCTION_CONFIG =
+      "distinctCountAggFunction";
+  private static final String DEFAULT_PERCENTILE_AGGREGATION_FUNCTION = "PERCENTILETDIGEST";
+  private static final String DEFAULT_DISTINCT_COUNT_AGGREGATION_FUNCTION = "DISTINCTCOUNT";
```
laxmanchekka, line 16: I think we should make DISTINCT_COUNT_HLL as the default.

aaron-steinfeld: I left it this way for backwards compatibility. Without any custom config, it'll retain identical behavior as before, but it's easily changed.

laxmanchekka (approved):

laxmanchekka (approved):

kotharironak (approved):

PR #145, opened 2022-06-23
feat: add CONTAINS_KEY_LIKE operator

GurtejSohi:
## Description
This PR adds `CONTAINS_KEY_LIKE` operator for doing a regex based search on map fields (on map key column). 

### Testing
Added unit test.

### Checklist:

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -357,7 +357,7 @@ class QueryRequestToPinotSQLConverter {
         literals.set(0, value.getValue().getString());
       } else {
         throw new IllegalArgumentException(
-            "Unsupported arguments for CONTAINS_KEY / CONTAINS_KEYVALUE operator");
+            "Unsupported arguments for CONTAINS_KEY / CONTAINS_KEYVALUE / CONTAINS_KEY_LIKE operator");
       }
     }
 
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -149,8 +149,13 @@ class QueryRequestToPinotSQLConverter {
               handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
           builder.append(operator);
           builder.append("(");
-          builder.append(
-              convertExpressionToString(filter.getLhs(), paramsBuilder, executionContext));
+          if (isMapColumnName(filter.getLhs())) {
+            // supports Like operator on map key column
+            builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
+          } else {
+            builder.append(
+                convertExpressionToString(filter.getLhs(), paramsBuilder, executionContext));
+          }
           builder.append(",");
           builder.append(convertExpressionToString(rhs, paramsBuilder, executionContext));
           builder.append(")");
@@ -440,5 +445,9 @@ class QueryRequestToPinotSQLConverter {
         break;
     }
     return ret;
+  }
+
+  private boolean isMapColumnName(Expression expression) {
+    return getLogicalColumnName(expression).map(viewDefinition::isMap).orElse(false);
   }
 }
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -149,13 +149,8 @@ class QueryRequestToPinotSQLConverter {
               handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
           builder.append(operator);
           builder.append("(");
-          if (isMapColumnName(filter.getLhs())) {
-            // supports Like operator on map key column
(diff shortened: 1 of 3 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -149,8 +149,13 @@ private String convertFilterToString(
               handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
           builder.append(operator);
           builder.append("(");
-          builder.append(
-              convertExpressionToString(filter.getLhs(), paramsBuilder, executionContext));
+          if (isMapColumnName(filter.getLhs())) {
+            // supports Like operator on map key column
+            builder.append(convertExpressionToMapKeyColumn(filter.getLhs()));
```
GurtejSohi, line 154: On second thought, this does a regex search only on the field's keys and not on its values. Should we do a regex search on the field's values as well here, or should we introduce new operators like - `KEY_LIKE` and `VALUE_LIKE` and leave the `LIKE` operator unchanged?

aaron-steinfeld: We already support operations on values via attribute expressions. We don't really want to introduce new operators, but the only potentially useful scenariosI could think of that we would be missing would be extensions of the `CONTAINS_KEY` operator - e.g. `NOT_CONTAINS_KEY` and `CONTAINS_KEY_LIKE`.

avinashkolluru: @GurtejSohi `CONTAINS_KEY_LIKE` is what i think we need

GurtejSohi: Removed these changes and added a new operator: `CONTAINS_KEY_LIKE `

GurtejSohi: > We don't really want to introduce new operators, but the only potentially useful scenariosI could think of that we would be missing would be extensions of the CONTAINS_KEY operator - e.g. NOT_CONTAINS_KEY and CONTAINS_KEY_LIKE.

Actually, we already have `NOT_CONTAINS_KEY` operator.

aaron-steinfeld: > Actually, we already have NOT_CONTAINS_KEY operator.

I believe we implemented at QS but haven't exposed it up to the platform API layer.

aaron-steinfeld (commented):
LGTM. Need to get integration tests fixed first, which is a pinot upgrade issue (ran into the same thing on my PR https://github.com/hypertrace/query-service/pull/143#issuecomment-1164902644 - hoping to return to it later today unless you get a chance first.

GurtejSohi:
> LGTM. Need to get integration tests fixed first, which is a pinot upgrade issue (ran into the same thing on my PR [#143 (comment)](https://github.com/hypertrace/query-service/pull/143#issuecomment-1164902644) - hoping to return to it later today unless you get a chance first.

@aaron-steinfeld I was looking into this today. As pointed by the logs [here](https://github.com/hypertrace/query-service/runs/7037155662?check_suite_focus=true#step:8:1713), [this check](https://github.com/hypertrace/query-service/blob/main/query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java#L148) is failing. While debugging, observed that [this map of kafka consumer group offsets](https://github.com/hypertrace/query-service/blob/main/query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java#L278) is empty, which implies that pinot is not consuming from the topics written to by view-generator (maybe some change in ingestion config key in the new version of pinot). Tried looking into it, but couldn't find the reason.

aaron-steinfeld:
> > LGTM. Need to get integration tests fixed first, which is a pinot upgrade issue (ran into the same thing on my PR [#143 (comment)](https://github.com/hypertrace/query-service/pull/143#issuecomment-1164902644) - hoping to return to it later today unless you get a chance first.
> 
> @aaron-steinfeld I was looking into this today. As pointed by the logs [here](https://github.com/hypertrace/query-service/runs/7037155662?check_suite_focus=true#step:8:1713), [this check](https://github.com/hypertrace/query-service/blob/main/query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java#L148) is failing. While debugging, observed that [this map of kafka consumer group offsets](https://github.com/hypertrace/query-service/blob/main/query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java#L278) is empty, which implies that pinot is not consuming from the topics written to by view-generator (maybe some change in ingestion config key in the new version of pinot). Tried looking into it, but couldn't find the reason.

Yes, I spent a few hours on this today. There's a lot going on, but the customer pinot service manager build wasn't updated to work with the current pinot version that it unpacks. One issue is its logging, but there seems to be a change in the consumer group behavior too. That said, it does actually ingest, it's just this check that's broken. Hoping @laxman-traceable or someone with more pinot and zookeeper familiarity can assist here.

laxmanchekka:
It can’t find that `""` consumer group. This is a change done during upgrade (0.7.0  to 0.10.0)
Earlier group name was empty string and doesn't allow us to configure.
Now, it allows us to configure. We are giving relevant names for each table during view creation. It's not  `""` empty string any more.

aaron-steinfeld (approved):

PR #155, opened 2022-07-21
feat: Added integration tests with postgres

saxenakshitiz:


query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPostgresQueriesTest.java
```diff
@@ -127,6 +130,7 @@ public static void setup() throws Exception {
   @AfterAll
   public static void shutdown() {
     LOG.info("Initiating shutdown");
+    IntegrationTestServerUtil.shutdownServices();
```
kotharironak, line 133: Do the same on Pinot tests too.

saxenakshitiz: Already added. Didn't resolve the issue.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
```diff
@@ -138,46 +141,87 @@ private String convertFilterToString(
...
+    return isSimpleAttributeExpression(filter.getLhs())
+        && filter.getRhs().getValueCase().equals(LITERAL)
+        && context.isBytesColumnType();
+  }
+
+  /** Handles value conversion of a bytes expression based */
+  private void handleConversionForBytesExpression(
+      String lhs, String operator, Expression rhs, StringBuilder builder, Builder paramsBuilder) {
+    Value value = rhs.getLiteral().getValue();
+    if (value.getValueType().equals(ValueType.STRING)) {
```
kotharironak, line 174: nit: can we move to a separate function 
`handleEqualAndNotEqualOperatorForByteExpression`?

kotharironak (approved):
overall looks fine. we can handle follow up
- nit for refactoring handling of equal or not equal operator
- running both test pinot + postgres together

PR #157, opened 2022-07-26
chore: Split postgres url in to parts.

saxenakshitiz:


query-service/src/integrationTest/resources/configs/query-service/postgres/application.conf
```diff
@@ -15,8 +15,12 @@ service.config = {
   clients = [
     {
       type = postgres
-      connectionString = "jdbc:postgresql://localhost:5432/traceable?user=postgres&password=postgres"
-      connectionString = ${?POSTGRES_CONNECT_STR}
+      host = localhost
```
kotharironak, line 18: For less envs + already prepared connection string, let's go with:
- connection string
- user
- password

What do you say?
```
connectionString = "jdbc:postgresql://localhost:5432/traceable
connectionString = ${?POSTGRES_CONNECT_STR}
user = postgres
user = ${?POSTGRES_USER}
password = postgres
password = ${?POSTGRES_PASSWROD}
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java
```diff
@@ -62,12 +62,36 @@ private RequestHandlerConfig(Config config) {
...
+    String host;
+    int port;
+    String databaseName;
+    String user;
+    String password;
 
     private RequestHandlerClientConfig(Config config) {
       this.type = config.getString(CONFIG_PATH_TYPE);
-      this.connectionString = config.getString(CONFIG_PATH_CONNECTION_STRING);
+      if (config.hasPath(CONFIG_PATH_CONNECTION_STRING)) {
```
kotharironak, line 80: Let's simplify this - connection string + user + password, so no if, those are must config.

query-service/src/integrationTest/resources/configs/query-service/postgres/application.conf
saxenakshitiz: Sure. However user and password are optional for pinot. Correct?

kotharironak: > Sure. However user and password are optional for pinot. Correct?

Yes, we only need this in Postgres for now.

saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java
saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java
```diff
@@ -62,12 +63,26 @@ private RequestHandlerConfig(Config config) {
...
     String type;
     String connectionString;
+    Optional<String> user;
+    Optional<String> password;
 
     private RequestHandlerClientConfig(Config config) {
       this.type = config.getString(CONFIG_PATH_TYPE);
       this.connectionString = config.getString(CONFIG_PATH_CONNECTION_STRING);
+      if (config.hasPath(CONFIG_PATH_USER)) {
+        this.user = Optional.of(config.getString(CONFIG_PATH_USER));
```
kotharironak, line 77: nit:
```
this.user = config.hasPath(CONFIG_PATH_USER) ? Optional.of(config.getString(CONFIG_PATH_USER)) : Optional.empty()
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresResultAnalyzer.java
```diff
@@ -1,75 +0,0 @@
...
-  }
-
-  /** For each selected attributes build the map of logical name to result index. */
-  static PostgresResultAnalyzer create(
-      ResultSet resultSet,
-      LinkedHashSet<String> selectedAttributes,
-      TableDefinition tableDefinition)
-      throws SQLException {
-    Map<String, Integer> physicalNameColIndexMap = new HashMap<>();
-    ResultSetMetaData metaData = resultSet.getMetaData();
```
kotharironak, line 41: Is our selection list of SQL queries in the same order as `columnMetadataSet` - https://github.com/hypertrace/query-service/blob/56613dafe1bf8eaa6419a26fcaf3f4c7a07755be/query-service-impl/src/main/java/org/hypertrace/core/query/service/ExecutionContext.java#L109 ?

saxenakshitiz: Yes, it is. The conversion from query to SQL is in same sequence. Therefore the result set returned contains data in same sequence.

kotharironak: > Yes, it is. The conversion from query to SQL is in same sequence. Therefore the result set returned contains data in same sequence.

I see we are using the selection list from the execution context only, so it will be in the same order.
```
Entry<String, Params> sql =
          request2PostgresSqlConverter.toSQL(
              executionContext, request, executionContext.getAllSelections());
```

kotharironak (approved):
overall, lgtm

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceConfig.java
saxenakshitiz: Will do it in follow up PR

PR #158, opened 2022-07-26
chore: remove the code not used in postgres query

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -478,77 +458,11 @@ && isEquals(tableColumnFilter.getValues(), queryFilter.getRhs())) {
...
-  private void handleSelection(
-      ResultSet resultSet, List<Builder> rowBuilderList, LinkedHashSet<String> selectedAttributes)
-      throws SQLException {
-    do {
-      // Find the index in the result's column for each selected attributes
-      PostgresResultAnalyzer resultAnalyzer =
-          PostgresResultAnalyzer.create(resultSet, selectedAttributes, tableDefinition);
-      Builder builder;
-      builder = Row.newBuilder();
+    while (resultSet.next()) {
```
kotharironak, line 463: Just a Note: Looks fine, just verify once that column order of the result set matches with the column order returned in query-request response metadata (As I remember, it's the same selection order, but let's cross-check once).

kotharironak (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
saxenakshitiz: I have validated that part

PR #162, opened 2022-08-05
fix: Remove NULL_STRING value type and use STRING value type

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -472,7 +472,8 @@ Observable<Row> convert(ResultSet resultSet) throws SQLException {
         for (int c = 1; c <= columnCount; c++) {
           String colVal = resultSet.getString(c);
```
aaron-steinfeld, line 473: Do we treat every result as a string? is that consistent with what we do in pinot?

aaron-steinfeld (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
aaron-steinfeld: Sounds like yes

kotharironak (approved):

PR #164, opened 2022-08-09
fix: Changes to handle conversion of array field to json array string…

saxenakshitiz:


kotharironak (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -470,13 +477,25 @@ Observable<Row> convert(ResultSet resultSet) throws SQLException {
...
+          int colType = metaData.getColumnType(c);
+          Value convertedColVal;
+          if (colType == Types.ARRAY) {
+            Array colVal = resultSet.getArray(c);
+            convertedColVal =
+                Value.newBuilder()
+                    .setValueType(ValueType.STRING)
+                    .setString(
+                        MAPPER.writeValueAsString(
+                            colVal != null ? colVal.getArray() : Collections.emptyList()))
```
kotharironak, line 489: In case of pinot, for `Collections.emptyList()` , it returns `["null"]`. We can add defaults in schema.

PR #166, opened 2022-08-11
chore: provide null value equivalent based on column type

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -493,7 +499,7 @@ Observable<Row> convert(ResultSet resultSet) {
             convertedColVal =
                 colVal != null
                     ? Value.newBuilder().setValueType(ValueType.STRING).setString(colVal).build()
-                    : NULL_VALUE;
+                    : getNullValueEquivalent(metaData.getColumnType(c));
```
aaron-steinfeld, line 502: I was a bit out of the loop on these conversations, I thought we were planning on setting the defaults in the db schema?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -64,8 +64,14 @@ public class PostgresBasedRequestHandler implements RequestHandler {
   private static final int DEFAULT_SLOW_QUERY_THRESHOLD_MS = 3000;
   private static final Set<Operator> GTE_OPERATORS = Set.of(Operator.GE, Operator.GT, Operator.EQ);
 
-  private static final Value NULL_VALUE =
+  private static final Value NULL_STRING_VALUE =
       Value.newBuilder().setValueType(ValueType.STRING).setString("null").build();
+  private static final Value NULL_INT_VALUE =
+      Value.newBuilder().setValueType(ValueType.STRING).setString("0").build();
```
aaron-steinfeld, line 70: The value type appears off here, but I'm guessing this is the equivalent of what happens in pinot? If that's the case, can we leave a comment explaining because without context it looks like a bug.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
kotharironak: @saxenakshitiz  Yeah, me too. That's what we discussed. Does this mean, we are going with query-service approach?
I think, we will still need that - https://github.com/traceableai/view-generation/. This will handle the problem only on response time. 

e.g What would happen on query request side?

kotharironak: If, we go with this path, I think, we will have to explore a few things on the query side. However, this can open up scenarios of differentiation between real null vs `null` as actual value. 

- if there is a filter on int type column: 
  ```
  select * from span-event-view where api_trace_count > 0
  ```
   Do we have to rewrite the query by adding `IS NOT NULL` condition? or it just work w/o that? 
   ```
    select * from span-event-view where api_trace_count IS NOT NULL and api_trace_count > 0 
    ```
 - For the string column, we will have to handle NULL
   ```
    select * from span-event-view where request_body IS NOT NULL
   ```

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
kotharironak: yes, currently, from query-service side, we are always returning string value, and conversation of an appropriate attribute.kind is handled at the gateway layer. we can add comments here.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
saxenakshitiz: > Do we have to rewrite the query by adding IS NOT NULL condition? or it just work w/o that?

In this case `IS NOT NULL` is not required. Postgres will handle it, and will not match null value. The only concern is whether `api_trace_count = 0` should change to `(api_trace_count IS NULL OR api_trace_count > 0)`.

saxenakshitiz: I find it simpler to keep changes in code rather than reflect it in DB. Will reduce storage requirement around storing null equivalent values.

kotharironak: Yes, keeping in code will reduce storage, and smaller row length, and avoid schema alteration. We need to see the impact on perf with defaults. I am just guessing,  should not be significant.

As in the above example - `api_trace_count = 0`, the pinot result will include those rows with the value `0`. So, yes, we will need `api_trace_count IS NULL OR api_trace_count > 0`. we will need to do the same for all other types, and strings need to handle separately. 

One more question, do we have to fully same as existing result - eg. would `avg(api_trace_count) where  api_trace_count IS NULL OR api_trace_count > 0` vs `avg(api_trace_count)` same? Does this matter?

For now, I think, to make it compatible with existing o/p that is on pinot default - https://docs.pinot.apache.org/basics/components/schema, it would be easier to have schema defaults.

kotharironak:
Have merged this one - https://github.com/Traceableai/view-generation/pull/239.

kotharironak:
Even after having defaults, we can still have `null` values in a column if they were explicitly inserted.
e.g
```
CREATE TABLE IF NOT EXISTS public."my-test-null"
(
    tenant_id text COLLATE pg_catalog."default" NOT NULL,
    user_score integer DEFAULT 0,
    user_name text COLLATE pg_catalog."default" DEFAULT 'null',
    seq_id bigserial
)

INSERT INTO  public."my-test-null" VALUES ('t1');
INSERT INTO  public."my-test-null" VALUES ('t1', null, null);

select * from  public."my-test-null";
```
Screenshot 2022-08-18 at 4 36 41 PM

So, for now, we need this fix as well along with default.

kotharironak (approved):

PR #167, opened 2022-08-12
chore: handle bytea field having default value

saxenakshitiz:


kotharironak:
As we are going with default for all the types including bytes (https://github.com/Traceableai/view-generation/pull/239), we need this change.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
```diff
@@ -195,18 +195,13 @@ private boolean handleConversionForNullOrEmptyBytesLiteral(
...
-          builder.append("IS NULL");
-          break;
-        case "!=":
-          builder.append("IS NOT NULL");
-          break;
-        default:
-          throw new IllegalArgumentException(
-              String.format(
-                  "Unsupported operator {%s} for bytes column with empty value", operator));
+      if (!operator.equals("=") && !operator.equals("!=")) {
```
kotharironak, line 198: `! ((operator.equals("=") || operator.equals("!="))` should we write this one?

kotharironak (commented):
lgtm

kotharironak (approved):

PR #168, opened 2022-08-13
feat: add support for array fields for postgres

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/TableDefinition.java
```diff
@@ -124,6 +132,8 @@ public static TableDefinition parse(
         spec = new PostgresColumnSpec(physName, ValueType.BYTES, false);
       } else if (tdigestFields.contains(physName)) {
         spec = new PostgresColumnSpec(physName, ValueType.STRING, true);
+      } else if (arrayFields.contains(physName)) {
```
kotharironak, line 135: Note: Currently, we have only `array<string>` in avdl files, so this will suffice for now. But, we will have to extend the same logic for any other array types (`array<double>`, array<int>`), etc.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
```diff
@@ -207,6 +211,63 @@ private boolean handleConversionForNullOrEmptyBytesLiteral(
...
+  private void handleConversionForArrayColumnExpression(
+      String lhs, String operator, Expression rhs, StringBuilder builder, Builder paramsBuilder) {
+    Value value = rhs.getLiteral().getValue();
+
+    if (handleConversionForNullOrEmptyArrayLiteral(lhs, operator, builder, value)) return;
+
+    builder.append(lhs);
+    // support only equals and IN operator
+    // both of them are handled as contains check
+    if (!operator.equals("=") && !operator.equals("IN")) {
```
kotharironak, line 230: We may have to handle other operators too. But, as basic can't we support `not in` and `!=`?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
```diff
@@ -207,6 +211,63 @@ private boolean handleConversionForNullOrEmptyBytesLiteral(
...
+      throw new IllegalArgumentException(
+          String.format(
+              "Unsupported operator {%s} for array column with non-empty value", operator));
+    } else {
+      // overlap operator for array
+      builder.append(" && ");
```
kotharironak, line 236: Can we add comments that `IN` and `EQ` are treated as contains in existing impl (say pinot), and so in postgres it translated to same using `&&` operator - https://www.postgresql.org/docs/current/functions-array.html?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/TableDefinition.java
saxenakshitiz: We will add support as and when needed

kotharironak: Lets have backlog ticket for this.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
```diff
@@ -207,6 +211,70 @@ private boolean handleConversionForNullOrEmptyBytesLiteral(
...
+    Value value = rhs.getLiteral().getValue();
+
+    if (handleConversionForNullOrEmptyArrayLiteral(lhs, operator, builder, value)) return;
+
+    // support only equals and IN operator
+    // both of them are handled as contains check to align with existing implementation
+    if (operator.equals("=")
+        || operator.equals("IN")
+        || operator.equals("!=")
+        || operator.equals("NOT IN")) {
```
kotharironak, line 232: In the same, backlog ticket, also add that other operators will be supported when required.

kotharironak (approved):
lgtm for string_array and `eq`, `in`, `not eq`, and for `not in` operators.

PR #169, opened 2022-08-17
feat: add connection pooling for postgres connections

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,32 +81,56 @@ public PostgresClient getPostgresClient(String clusterName) {
...
+          "Trying to create a Postgres client connected to postgres server using url: {}, user: {}, password: {}",
+          url,
+          user,
+          password);
+      ConnectionFactory connectionFactory = new DriverManagerConnectionFactory(url, user, password);
+      PoolableConnectionFactory poolableConnectionFactory =
+          new PoolableConnectionFactory(connectionFactory, null);
+      GenericObjectPool<PoolableConnection> connectionPool =
+          new GenericObjectPool<>(poolableConnectionFactory);
+      connectionPool.setMaxTotal(50);
```
saxenakshitiz, line 104: Once we are fine with current approach, I will move these values to config

kotharironak: yes, we will need that. 

Prior, we were having one connection/handler. With this, we will have a pool of connections/handler. This will require an appropriate config for PgPool + max_connection setting on the server.

What are the defaults values?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,32 +81,56 @@ public PostgresClient getPostgresClient(String clusterName) {
...
+    public Observable<Row> executeQuery(String statement, Params params) throws SQLException {
       String resolvedStatement = resolveStatement(statement, params);
-      return connection.prepareStatement(resolvedStatement);
+      try (Connection connection = dataSource.getConnection();
+          PreparedStatement preparedStatement = connection.prepareStatement(resolvedStatement);
+          ResultSet resultSet = preparedStatement.executeQuery()) {
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("Query results: [ {} ]", resultSet);
+        }
+        return convert(resultSet);
```
kotharironak, line 132: Shall we keep this as is? Result conversion as separate. As part of this class, let's have client related.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,32 +81,56 @@ public PostgresClient getPostgresClient(String clusterName) {
...
-      LOG.debug(
-          "Trying to create a Postgres client connected to postgres server using url: {}", url);
-      this.connection = DriverManager.getConnection(url);
+      String url = clientConfig.getConnectionString();
+      this.dataSource = createPooledDataSource(url, user, password);
     }
 
-    private PostgresClient(Connection connection) {
-      this.connection = connection;
+    private DataSource createPooledDataSource(String url, String user, String password) {
```
aaron-steinfeld, line 93: This is getting really messy really quickly. Why do we need this here and not in doc store for example? Should the postgres client building code be in doc store and shared here?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,32 +81,56 @@ public PostgresClient getPostgresClient(String clusterName) {
...
+      LOG.debug(
+          "Trying to create a Postgres client connected to postgres server using url: {}, user: {}, password: {}",
```
aaron-steinfeld, line 95: Debug logs have more flexibility, but is there really value in logging the password here?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,32 +81,56 @@ public PostgresClient getPostgresClient(String clusterName) {
...
 
-    private PreparedStatement buildPreparedStatement(String statement, Params params)
-        throws SQLException {
+    public Observable<Row> executeQuery(String statement, Params params) throws SQLException {
       String resolvedStatement = resolveStatement(statement, params);
-      return connection.prepareStatement(resolvedStatement);
+      try (Connection connection = dataSource.getConnection();
+          PreparedStatement preparedStatement = connection.prepareStatement(resolvedStatement);
+          ResultSet resultSet = preparedStatement.executeQuery()) {
+        if (LOG.isDebugEnabled()) {
```
aaron-steinfeld, line 129: nit: unnecessary if condition, that's what LOG.debug is there for.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -131,5 +178,42 @@ static String resolveStatement(String query, Params params) {
     private static String getStringParam(String value) {
       return "'" + value.replace("'", "''") + "'";
     }
+
+    @SneakyThrows
+    Observable<Row> convert(ResultSet resultSet) {
+      List<Row.Builder> rowBuilderList = new ArrayList<>();
+      while (resultSet.next()) {
```
aaron-steinfeld, line 185: Seems like this would be cleaner if streamed and immutable.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -131,5 +178,42 @@ static String resolveStatement(String query, Params params) {
...
+        Row.Builder builder = Row.newBuilder();
+        rowBuilderList.add(builder);
```
aaron-steinfeld, line 187: but if not, add at the end and create a list of rows rather than a list of row builders.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -131,5 +178,42 @@ static String resolveStatement(String query, Params params) {
...
+              convertedColVal =
+                  colVal != null
+                      ? Value.newBuilder().setValueType(ValueType.STRING).setString(colVal).build()
+                      : NULL_VALUE;
+            }
+            builder.addColumn(convertedColVal);
+          }
+        }
+      }
+      return Observable.fromIterable(rowBuilderList)
```
aaron-steinfeld, line 214: I love observables as much (more) as the next guy, but this observable isn't being used for anything async, so what's the purpose? If there's a reason to use Rx, should start with the actual db call and make the whole thing observable rather than just doing everything sync and wrapping the result.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -392,27 +378,23 @@ public Observable<Row> handleRequest(
...
+      return rowObservable.doOnComplete(
+          () -> {
+            long requestTimeMs = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS);
+            if (requestTimeMs > slowQueryThreshold) {
+              try {
+                LOG.warn(
+                    "Query Execution time: {} ms, sqlQuery: {}, queryRequest: {}, executionStats: {}",
+                    requestTimeMs,
+                    sql.getKey(),
+                    protoJsonPrinter.print(request),
```
aaron-steinfeld, line 391: So _this_ is where you want that if check. Your arguments are doing the work rather than the logger. Given that this is only triggered on threshold and log level is warn, may not be worth doing, but thats the difference from the other places.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: We need to do conversion before result set is closed. We can move convert method to a separate class.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Document store and query service are different.
Document store handles query for data stored within a json document.
Query service handles query for data stored in multiple column within a table as of now.
In short term, they will be separate impl.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Will remove the password

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
aaron-steinfeld: Agreed, but just to clarify I meant doc store the repo rather than doc store the lib. Both the doc store lib and query service talk to postgres, so using a common postgres connection/client lib between them would make sense, no? (this is the bit that I was suggesting could be stored in doc store the repo)

saxenakshitiz: Will move to postgres repo and use it from both places. However it will be done as a followup.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
saxenakshitiz: Got it :-)

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,78 +62,60 @@ public PostgresClient getPostgresClient(String clusterName) {
...
-      params.getLongParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-      params.getDoubleParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-      params.getFloatParams().forEach((i, p) -> parameters[i] = String.valueOf(p));
-
-      StringBuilder sb = new StringBuilder();
-      for (int i = 0; i < queryParts.length; i++) {
-        sb.append(queryParts[i]);
-        sb.append(parameters[i] != null ? parameters[i] : "");
+    private int getPercentOf(int maxConnections, int percent) {
+      int value = (maxConnections * 100) / percent;
```
kotharironak, line 113: For `maxConnection` as `5` this will be `25` based on above method call - `getPercentOf(maxConnections, 20)`.

Shall we do with ?
```
Math.floor(maxConnections * percent/100) 
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -58,78 +62,60 @@ public PostgresClient getPostgresClient(String clusterName) {
...
+      // min value is always 1
+      if (value < 1) {
```
kotharironak, line 115: nit: `return (value < 1) ? 1 : value`

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Fixed the calculation. Don't need to use floor function.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Done

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: result set is not streamable

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -110,12 +110,8 @@ public Connection getConnection() throws SQLException {
...
 
     private int getPercentOf(int maxConnections, int percent) {
-      int value = (maxConnections * 100) / percent;
-      // min value is always 1
-      if (value < 1) {
-        return 1;
-      }
-      return value;
+      int value = (maxConnections * percent) / 100;
+      return value > 1 ? value : 0;
```
kotharironak, line 114: This will be `return value > 1 ? value : 1;` ?

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: Caller requires an observable to be returned. However you are right current implementation is not async.

kotharironak (approved):

PR #172, opened 2022-08-24
feat: create new connection on connection validation error in postgres

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
```diff
@@ -62,57 +59,98 @@ public PostgresClient getPostgresClient(String clusterName) {
...
+      }
     }
 
-    public Connection getConnection() throws SQLException {
-      return dataSource.getConnection();
+    private synchronized void newConnection() throws SQLException {
+      int attempts = 0;
+      while (attempts < maxConnectionAttempts) {
+        try {
+          ++count;
```
surajpuvvada, line 118: what is count used for ?

kotharironak, line 118: Same comment on as the previous PR.

kotharironak (commented):
lgtm

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresClientFactory.java
saxenakshitiz: it is sort of id for connection

saxenakshitiz: done

kotharironak (approved):

PR #177, opened 2022-11-07
chore: make like operator query to be case insensitive

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/converters/DefaultColumnRequestConverter.java
```diff
@@ -33,7 +33,7 @@
 class DefaultColumnRequestConverter implements ColumnRequestConverter {
 
   private static final String QUESTION_MARK = "?";
-  private static final String REGEX_OPERATOR = "~";
+  private static final String REGEX_OPERATOR = "~*";
```
aaron-steinfeld, line 36: Not that it's relevant, but I wouldn't expect this to be case insensitive. What _is_ relevant is - what is our pinot implementation? They need to match.

saxenakshitiz: The search is case insensitive in Pinot. Just aligning the implementation.

aaron-steinfeld (approved):

PR #178, opened 2022-12-15
Added support for accurate distinct count on attributes through config

singhalprerana:
Existing config supports configuring the distinct count aggregation function for a given scope for all attributes. But we might need accurate distinct count on some attributes if the default is set to the approximate function. Defined a way to configure such attributes.

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
@@ -1,10 +1,11 @@
 package org.hypertrace.core.query.service.pinot.converters;
 
-import com.google.common.collect.ImmutableSet;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import java.util.Collections;
-import java.util.Set;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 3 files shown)


aaron-steinfeld:
I'd love to dive in to the use case a bit if you want to set up a call - remember the reason this went in in the first place was with enough the performance of exact distinct count wasn't satisfactory. So will that be better for some columns rather than others? If not, then we're just opening up a config to let us shoot ourselves in the foot.

singhalprerana:
> I'd love to dive in to the use case a bit if you want to set up a call - remember the reason this went in in the first place was with enough the performance of exact distinct count wasn't satisfactory. So will that be better for some columns rather than others? If not, then we're just opening up a config to let us shoot ourselves in the foot.

@aaron-steinfeld So the context is we have many mismatches in grouping queries which surface to UI, even for mall numbers.. In certain cases, this might be critical, e.g. distinct number of APIs or domains involved in threat activity - https://traceableai.atlassian.net/browse/ENG-20719.. We want accurate counts in queries like grouping by domains and getting distinct count of APIs.
that’s why we wanted to allow accurate counts for certain fields which we can control via helm instead of exposing this function through GQL API..

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
```diff
@@ -1,22 +1,30 @@
...
 
 @Value
-@AllArgsConstructor
+@AllArgsConstructor(access = AccessLevel.PACKAGE)
 public class PinotFunctionConverterConfig {
 
   private static final String PERCENTILE_AGGREGATION_FUNCTION_CONFIG = "percentileAggFunction";
   private static final String DISTINCT_COUNT_AGGREGATION_FUNCTION_CONFIG =
       "distinctCountAggFunction";
+  private static final String ARGS_FOR_ACCURATE_DISTINCT_COUNT_AGG_CONFIG =
```
aaron-steinfeld, line 20: This isn't super generic, since the distinct count function it self is configurable. Can we update the config so we basically have a default distinct count an an override map? (either function to list of column names or column name to function)

singhalprerana: done

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
```diff
@@ -30,9 +38,24 @@ public PinotFunctionConverterConfig(Config config) {
     } else {
       this.distinctCountAggregationFunction = DEFAULT_DISTINCT_COUNT_AGGREGATION_FUNCTION;
     }
+    if (config.hasPath(DISTINCT_COUNT_AGGREGATION_OVERRIDES)) {
+      Config overridesConfig = config.getConfig(DISTINCT_COUNT_AGGREGATION_OVERRIDES);
+      this.distinctCountAggOverrides =
+          overridesConfig.entrySet().stream()
```
aaron-steinfeld, line 44: nit: keySet, not using the whole entry

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
```diff
@@ -30,9 +38,24 @@ public PinotFunctionConverterConfig(Config config) {
...
+    } else {
+      this.distinctCountAggOverrides = Collections.emptyMap();
+    }
   }
 
   public PinotFunctionConverterConfig() {
     this(ConfigFactory.empty());
   }
+
+  public String getDistinctCountFunction(String arg) {
```
aaron-steinfeld, line 57: nit rename arg to what the string is supposed to represent (columnName)

aaron-steinfeld (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
singhalprerana: this is config object - no keySet available, only entrySet..

aaron-steinfeld: ah - `ConfigObject` has keySet (it implements map), `Config` does not. So `.root().keySet()` would work. FWIW, the two are equivalent for a one level config but I would say keySet version better aligns with your intention - getting the immediate keys, rather than what entrySet does (recurses into all children and produces path expressions mapped to any descendent value).

PR #181, opened 2023-03-12
adding time based filter for aggregation views

kotharironak:
## Description
As part of Postgres implementation, it adds the support of choosing a request handler based on query request time range.
In Postgres, there was a requirement to have multiple levels of aggregations for time series metrics data - 30 secs, 5 mins, and so on. This requires different table selections based on the query request time range. So, as part of Postgres implementation, it adds the support of choosing a request handler based on the query request time range.

### Testing
Added the required unit tests

### Checklist:


query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -196,6 +205,13 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
       cost = tableDefinition.getTimeGranularityMillis() / (Long.MAX_VALUE * 2D);
     }
 
+    long requestEndTime = getRequestEndTime(request.getFilter());
+    Duration requestDuration = Duration.ofMillis(requestEndTime - requestStartTime);
+
+    if (minRequestDuration != Duration.ZERO
```
saxenakshitiz, line 211: Why do we need to check for Duration Zero explicitly. Single compare check as below should be good enough. We will anyways have order defined in query service config, and handler will higher minDuration will be defined first.

kotharironak: We can do that. I was keeping the default handler first in the list. 

With this, we have to keep the handler order from highest to lowest minRequestDuration. Logically, this is easy to maintain, addressing the comments.

e.g if there are three handlers as described in test cases, we will have to list them in the following order.
- backend-traces-from-bare-span-event-view-5min-aggr-handler
    - minRequestDuration : 12 hrs
- backend-traces-from-bare-span-event-view-30secs-aggr-handler
    - minRequestDuration : 3 hrs
- backend-traces-from-bare-span-event-view-aggr-handler 
    - minRequestDuration : 0 hrs

kotharironak: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/postgres/PostgresBasedRequestHandler.java
```diff
@@ -196,6 +205,13 @@ public QueryCost canHandle(QueryRequest request, ExecutionContext executionConte
       cost = tableDefinition.getTimeGranularityMillis() / (Long.MAX_VALUE * 2D);
     }
 
+    long requestEndTime = getRequestEndTime(request.getFilter());
+    Duration requestDuration = Duration.ofMillis(requestEndTime - requestStartTime);
+
+    // check: requestDuration < minRequestDuration
+    if (!(requestDuration.compareTo(minRequestDuration) < 0)) {
```
saxenakshitiz, line 212:
```suggestion
    if (requestDuration.compareTo(minRequestDuration) >= 0) {
```

kotharironak: As we are selecting a handler based on minimum cost, here, would like to reduce the cost of the hander that can server has met.

In this `requestDuration.compareTo(minRequestDuration) >= 0`, I don't want to change anything.

kotharironak: Oh! okey. I will change the boolean condition.

kotharironak: Done, misunderstood with the ordering of handler.

saxenakshitiz (approved):

avinashkolluru (approved):

surajpuvvada (approved):
kshitiz has already reviewed

PR #186, opened 2023-06-13
feat: add support for utilising text indexes if exits for like operator in pinot based handlers

kotharironak:
## Description
This PR add support for utilising text indexes for LIKE operator for pinot based request handler if the index for those column is enabled. So, the string column with index will translated to 
- text_match(column, /\<value\>/)

### Testing
- Added the request conversation test.

### Checklist:

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -151,8 +151,7 @@ class QueryRequestToPinotSQLConverter {
            */
           operator = handleLikeOperatorConversion(filter.getLhs());
           Expression rhs =
-              handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
-          rhs = postProcessValueConversionForLikeOperator(operator, rhs);
+              handleValueConversionForLikeArgument(operator, filter.getLhs(), filter.getRhs());
 
           builder.append(operator);
           builder.append("(");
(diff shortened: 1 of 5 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -276,6 +283,28 @@ private String convertOperatorToString(Operator operator) {
...
+    Optional<String> logicalColumnName = getLogicalColumnName(expression);
+    if (logicalColumnName.isPresent()
+        && isSimpleAttributeExpression(expression)
+        && viewDefinition.hasTextIndex(logicalColumnName.get())) {
+      return TEXT_MATCH_OPERATOR;
+    }
+    return REGEX_OPERATOR;
+  }
+
+  private Expression prefixValueForLikeOperator(String likeOperatorStr, Expression rhsExpression) {
```
kotharironak, line 296: For TEXT_MATCH regex query needs string to be prefixed and suffixed with '/'.

- https://docs.pinot.apache.org/basics/indexing/text-search-support

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -144,9 +145,15 @@ private String convertFilterToString(
...
         case LIKE:
-          // The like operation in PQL looks like `regexp_like(lhs, rhs)`
+          /**
+           * If the text index is not enabled on lhs expression, - the pql looks like
+           * `regexp_like(lhs, rhs)` else - the pql looks like `text_match(lhs, rhs)`
+           */
+          operator = handleLikeOperatorConversion(filter.getLhs());
           Expression rhs =
               handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
+          rhs = postProcessValueConversionForLikeOperator(operator, rhs);
```
aaron-steinfeld, line 155: nit: rather than taking multiple steps to calculate the RHS in the top level, could you do something like
```java
          operator = handleLikeOperatorConversion(filter.getLhs());
          Expression rhs =
              handleValueConversionForLikeArgument(filter.getLhs(), filter.getRhs());
```

and then have `handleValueConversionForLikeArgument` do the delegation.

kotharironak: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/ViewDefinition.java
```diff
@@ -120,6 +128,11 @@ public static ViewDefinition parse(Config config, String tenantColumnName) {
         spec.addColumnName(physName);
         spec.setType(ValueType.STRING);
       }
+
+      if (textIndexFields.contains(physName)) {
+        spec.setTextIndex();
```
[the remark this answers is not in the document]
kotharironak: We are maintaining map of logical column names to PinotColumnSpec. The physical columnName is already added above before this lines #128, #129 as last `else` block.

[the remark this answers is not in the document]
aaron-steinfeld: ah got it, the code fold got me again! The spec here is a column spec (while I had thought it was a spec for the whole table, since we were adding rather than setting a column name on line 128)

aaron-steinfeld (approved):

PR #188, opened 2023-07-07
feat: extend the support for distinct count for multi-value

kotharironak:
## Description
As part of this PR, we extend of support for using different aggregation function for distinct count query for multi value column.

Currently, we do support the same for single value field via below config at request handler level:
```
distinctCountAggFunction = DISTINCTCOUNTHLL
distinctCountAggOverrides = {
 "attribute_id_1" = DISTINCTCOUNT
}
```
It tells that, for all the single value attribute except `attribute_id_1` of the request handler, it will use `DISTINCTCOUNTHLL` function.

We are extending above as below:
```
distinctCountAggFunction = DISTINCTCOUNTHLL
distinctCountMvAggFunction = DISTINCTCOUNTHLLMV
distinctCountAggOverrides = {
 "attribute_id_1" = DISTINCTCOUNT
"attribute_mv_1" = DISTINCTCOUNT
}
```
The above config tells that,
- for all multi-value attribute except `attribute_id_1`, it will `DISTINCTCOUNTHLLMV` for distinct count
- for all single value attribute except `attribute_mv_1, it will use `DISTINCTCOUNTHLL` for distinct count

### Testing
Have added unit test.. 

### Checklist:
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
@@ -20,7 +20,7 @@ import org.hypertrace.core.query.service.api.ValueType;
 
 public class PinotFunctionConverter {
   private static final String PINOT_CONCAT_FUNCTION = "CONCATSKIPNULL";
-  private static final String QUERY_FUNCTION_DISTINCTCOUNT_MV_INTERNAL = "DISTINCTCOUNTMV";
+  private static final String PINOT_DISTINCT_COUNT_MV_FUNCTION = "DISTINCTCOUNTMV";
 
   private static final String DEFAULT_AVG_RATE_SIZE = "PT1S";
   private final PinotFunctionConverterConfig config;
@@ -48,7 +48,7 @@ public class PinotFunctionConverter {
         return this.functionToStringForDistinctCount(function, argumentConverter);
       case QUERY_FUNCTION_CONCAT:
         return this.functionToString(this.toPinotConcat(function), argumentConverter);
-      case QUERY_FUNCTION_DISTINCTCOUNT_MV_INTERNAL:
+      case PINOT_DISTINCT_COUNT_MV_FUNCTION:
         return this.functionToStringForDistinctCountMv(function, argumentConverter);
       case QUERY_FUNCTION_AVGRATE:
         // AVGRATE not supported directly in Pinot. So AVG_RATE is computed by summing over all
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
@@ -78,13 +78,13 @@ public class PinotFunctionConverter {
   private String functionToStringForDistinctCount(
       Function function, java.util.function.Function<Expression, String> argumentConverter) {
     String columnName = argumentConverter.apply(function.getArgumentsList().get(0));
-    return this.config.getDistinctCountFunction(columnName) + "(" + columnName + ")";
+    return String.format("%s(%s)", this.config.getDistinctCountFunction(columnName), columnName);
   }
 
   private String functionToStringForDistinctCountMv(
       Function function, java.util.function.Function<Expression, String> argumentConverter) {
     String columnName = argumentConverter.apply(function.getArgumentsList().get(0));
-    return this.config.getDistinctCountMvFunction(columnName) + "(" + columnName + ")";
+    return String.format("%s(%s)", this.config.getDistinctCountMvFunction(columnName), columnName);
   }
 
   private String functionToStringForAvgRate(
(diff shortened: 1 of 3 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
```diff
@@ -18,12 +18,16 @@ public class PinotFunctionConverterConfig {
   private static final String PERCENTILE_AGGREGATION_FUNCTION_CONFIG = "percentileAggFunction";
   private static final String DISTINCT_COUNT_AGGREGATION_FUNCTION_CONFIG =
       "distinctCountAggFunction";
+  private static final String DISTINCT_COUNT_MV_AGGREGATION_FUNCTION_CONFIG =
+      "distinctCountMvAggFunction";
```
kotharironak, line 22: I did consider the approach of re-using existing configuration `distinctCountAggFunction`.

```
distinctCountAggFunction = DISTINCTCOUNTHLL
distinctCountAggOverrides = {
 "attribute_id_1" = DISTINCTCOUNT
}
```

We can consider above config as follow:
- for all single value attributes except `attribute_id_1`, use `DISTINCTCOUNTHLL`
- for all multi-value attributes use `DISTINCTCOUNTHLLMV`

However, the above approach will need internal mapping for corresponding multi-value function. In the above case:
```
multi-value-disctinct-count-func-map = {
"DISTINCTCOUNTHLL" : "DISTINCTCOUNTHLLMV"
}
```
So, for any new supported function, it will need a code change. So, thought of extending the config it self.

laxmanchekka (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -78,6 +81,12 @@ private String functionToStringForDistinctCount(
     return this.config.getDistinctCountFunction(columnName) + "(" + columnName + ")";
   }
 
+  private String functionToStringForDistinctCountMv(
+      Function function, java.util.function.Function<Expression, String> argumentConverter) {
+    String columnName = argumentConverter.apply(function.getArgumentsList().get(0));
+    return this.config.getDistinctCountMvFunction(columnName) + "(" + columnName + ")";
```
skjindal93, line 87: Can we use `String.format` to generate the query, just like in `functionToStringForAvgRate`?

mihirgt:
@skjindal93 Can you review again, I have fixed the review comment.

laxmanchekka (approved):

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverter.java
```diff
@@ -47,6 +48,8 @@ public String convert(
         return this.functionToStringForDistinctCount(function, argumentConverter);
       case QUERY_FUNCTION_CONCAT:
         return this.functionToString(this.toPinotConcat(function), argumentConverter);
+      case PINOT_DISTINCT_COUNT_MV_FUNCTION:
```
satish-mittal, line 51: nit: this constant can go to `QueryFunctionConstants` class, similar to QUERY_FUNCTION_DISTINCTCOUNT

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/converters/PinotFunctionConverterConfig.java
```diff
@@ -58,4 +68,9 @@ public String getDistinctCountFunction(String arg) {
     return Optional.ofNullable(distinctCountAggOverrides.get(arg))
         .orElse(distinctCountAggregationFunction);
   }
+
+  public String getDistinctCountMvFunction(String arg) {
+    return Optional.ofNullable(distinctCountAggOverrides.get(arg))
```
satish-mittal, line 73: Given that we have introduced separate configs for Agg function (SV vs MV), we should extend the same to overrides as well. Instead of reusing `distinctCountAggOverrides` for both Sv and Mv, we can have 2 separate overrides config: `distinctCountAggOverrides` and `distinctCountMvAggOverrides`. Will be error-free and easier to understand.

kotharironak: The idea here is to have overrides at the handler level based on the data type, whether it's a single value field or a multi-value field. The `distinctCountAggOverrides` allows for overriding specific attributes, which can be of any data type as it maps an attribute to a function.

satish-mittal: Single override will also work. Just that the onus of configuring correct flavour of function (SV vs MV) is on user. 
```
distinctCountAggOverrides = {
 "attr1" = DISTINCTCOUNT
"attr2" = DISTINCTCOUNT
}
```
If suppose attr2 is Mv, then the only valid alternative is DISTINCTCOUNTMV. Backend can only validate if there is some attribute metadata (SV vs MV) available (not sure about it).

satish-mittal (approved):

skjindal93 (approved):

PR #190, opened 2023-07-14
fix: adds values filter for map column

kotharironak:
## Description
This pull request introduces an additional value filter for map columns, aiming to enhance performance by utilizing the underlying inverted index on the `__keys` and `__values` columns of the Pinot implementation.

This optimisation is specifically applied to queries like `mapValue(tags__keys, 'flags', tags__values) = '0'`.

This approach is the same as the one followed in the deprecated `CONTAINS_KEYVALUE`.
Ref: https://github.com/hypertrace/query-service/blob/b89ebf416e876959586d47f8e67f126be239c7d5/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java#L181

### Testing
Modified the existing test case.

### Checklist:

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -203,7 +203,7 @@ class QueryRequestToPinotSQLConverter {
           break;
         default:
           rhs = handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
-          applyValueFilterForMapColumn(filter, paramsBuilder, executionContext, builder, rhs);
+          applyValuesFilterForMapColumn(filter, paramsBuilder, executionContext, builder, rhs);
           builder.append(
               convertExpressionToString(filter.getLhs(), paramsBuilder, executionContext));
           builder.append(" ");
@@ -225,7 +225,7 @@ class QueryRequestToPinotSQLConverter {
    * CONTAINS_KEYVALUE, we also added the "__values" filter. By including this additional filter, if
    * there is an inverted index on a map field, it can be utilized to improve performance.
    */
-  private void applyValueFilterForMapColumn(
+  private void applyValuesFilterForMapColumn(
       Filter filter,
       Builder paramsBuilder,
       ExecutionContext executionContext,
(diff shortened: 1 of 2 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -203,6 +203,12 @@ private String convertFilterToString(
           break;
         default:
           rhs = handleValueConversionForLiteralExpression(filter.getLhs(), filter.getRhs());
+
+          Optional<String> valueFilter =
+              getValuesFilterForMapColumn(
+                  filter.getLhs(), rhs, filter.getOperator(), paramsBuilder, executionContext);
+          valueFilter.stream().forEach(str -> builder.append(str));
```
satish-mittal, line 210:
```suggestion
          valueFilter.ifPresent(builder::append);
```
since there is no collection within Optional

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -214,6 +220,33 @@ private String convertFilterToString(
...
+      Expression lhs,
+      Expression rhs,
+      Operator operator,
+      Builder paramsBuilder,
+      ExecutionContext executionContext) {
+    if (isAttributeExpressionWithSubpath(lhs) && operator.equals(Operator.EQ)) {
+      String valCol = convertExpressionToMapValueColumn(lhs);
+      return Optional.of(
+          String.format(
+              "%s %s %s AND ",
```
satish-mittal, line 243: nit: we can take `AND ` outside this function, so that it can be reused elsewhere as a terminal filter condition (i.e. when there is no other filter after it).

kotharironak: Hmm, ok.

kotharironak: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
kotharironak: Done.

satish-mittal (approved):

skjindal93 (approved):

PR #170, opened 2022-08-22
fix: add support for boolean array fields for pinot

SrikarMannepalli:
## Description
This PR adds support for queries involving boolean array fields for Pinot.

Sample failing query  - 
```
{
    "filter": {
        "operator": "AND",
        "childFilter": [
            {
                "lhs": {
                    "attribute_expression": {
                        "attributeId": "API.isExternal",
                        "alias": "isExternal"
                    }
                },
                "rhs": {
                    "literal": {
                        "value": {
                            "valueType": "BOOLEAN_ARRAY",
                            "boolean_array": [false]
                        }
                    }
                },
                "operator":"IN"
            }
        ]
    },
    "limit": 1,
    "selection": [
        {
            "attribute_expression": {
                "attributeId": "API.isExternal",
                "alias": "isExternal"
            }
        }
    ]
}
```

### Testing
Added new unit tests for all new code, and ran all the existing unit tests.

### Checklist:
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #193, opened 2023-08-31
fix: add support for boolean values in IN clause

kotharironak:
## Description
Currently, query with boolean_array in `IN` filter clause is not supported. As part of this PR, we enhance that support for pinot's implementation.

In pinot, both the below are valid query:
```
-- value as boolean
select api_is_external, api_id from spanEventView
where start_time_millis > 1693470600000 and start_time_millis < 1693474200000
and is_external IN (true)
limit 10
```

```
-- value as string
select api_is_external, api_id from bareSpanEventView
where start_time_millis > 1693470600000 and start_time_millis < 1693474200000
and is_external IN ('true')
limit 10
```
So, as part of this PR, we are taking first approach of `value as boolean`. This way, it can support a column whose type is boolean in pinot as well as string column that were used for boolean type.

### Testing
Added test

### Checklist:

### DIFF
--- a/owasp-suppressions.xml
+++ b/owasp-suppressions.xml
@@ -18,7 +18,7 @@
     <cve>CVE-2020-13956</cve>
   </suppress>
 
-  <suppress until="2023-07-30Z">
+  <suppress until="2023-09-30Z">
     <notes><![CDATA[
    file name: zookeeper-api-1.2.0.jar
    ]]></notes>
@@ -28,7 +28,7 @@
     <cve>CVE-2019-0201</cve>
   </suppress>
 
-  <suppress until="2023-07-30Z">
+  <suppress until="2023-09-30Z">
     <notes><![CDATA[
    file name: commons-compiler-3.1.9.jar
    https://github.com/janino-compiler/janino/issues/201
@@ -37,7 +37,7 @@
     <cve>CVE-2023-33546</cve>
   </suppress>
 
-  <suppress until="2023-07-30Z">
+  <suppress until="2023-09-30Z">
     <notes><![CDATA[
     Same series of issues that also impacts jackson, vendor disputed and not currently fixed
     https://github.com/janino-compiler/janino/issues/201
@@ -46,7 +46,7 @@
     <cve>CVE-2023-33546</cve>
   </suppress>
 
-  <suppress until="2023-07-30Z">
+  <suppress until="2023-09-30Z">
     <notes><![CDATA[
      file name: jackson-databind-2.14.2.jar
      This is currently disputed.
--- a/owasp-suppressions.xml
+++ b/owasp-suppressions.xml
@@ -27,25 +27,6 @@
     <cve>CVE-2018-8012</cve>
     <cve>CVE-2019-0201</cve>
   </suppress>
-
-  <suppress until="2023-09-30Z">
-    <notes><![CDATA[
-   file name: commons-compiler-3.1.9.jar
-   https://github.com/janino-compiler/janino/issues/201
-   ]]></notes>
-    <packageUrl regex="true">^pkg:maven/org\.codehaus\.janino/commons\-compiler@.*$</packageUrl>
-    <cve>CVE-2023-33546</cve>
-  </suppress>
-
-  <suppress until="2023-09-30Z">
-    <notes><![CDATA[
-    Same series of issues that also impacts jackson, vendor disputed and not currently fixed
-    https://github.com/janino-compiler/janino/issues/201
-   ]]></notes>
-    <packageUrl regex="true">^pkg:maven/org\.codehaus\.janino/.*@.*$</packageUrl>
-    <cve>CVE-2023-33546</cve>
-  </suppress>
-
   <suppress until="2023-09-30Z">
     <notes><![CDATA[
      file name: jackson-databind-2.14.2.jar
@@ -53,5 +34,12 @@
      ]]></notes>
     <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson\-databind@.*$</packageUrl>
     <cve>CVE-2023-35116</cve>
+  </suppress>
+  <suppress until="2023-09-30Z">
+    <notes><![CDATA[
+   file name: netty-handler-4.1.94.Final.jar
+   ]]></notes>
+    <packageUrl regex="true">^pkg:maven/io\.netty/netty\-handler@.*$</packageUrl>
+    <vulnerabilityName>CVE-2023-4586</vulnerabilityName>
   </suppress>
 </suppressions>
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
@@ -79,6 +79,8 @@ public class Params {
         + doubleParams
         + ", byteStringParams="
         + byteStringParams
+        + ", booleanParams="
+        + booleanParams
         + '}';
   }
 
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
@@ -17,7 +17,6 @@ public class Params {
   private final Map<Integer, Float> floatParams;
   private final Map<Integer, Double> doubleParams;
   private final Map<Integer, ByteString> byteStringParams;
-  private final Map<Integer, Boolean> booleanParams;
 
   private Params(
       Map<Integer, Integer> integerParams,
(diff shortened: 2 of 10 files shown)


owasp-suppressions.xml
```diff
@@ -28,7 +28,7 @@
     <cve>CVE-2019-0201</cve>
   </suppress>
 
-  <suppress until="2023-07-30Z">
+  <suppress until="2023-09-30Z">
```
kotharironak, line 31: I am trying it out with latest 3.1.10.

kotharironak: Fixed this.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
```diff
@@ -57,6 +60,10 @@ public Map<Integer, ByteString> getByteStringParams() {
     return byteStringParams;
   }
 
+  public Map<Integer, Boolean> getBooleanParams() {
+    return booleanParams;
+  }
+
   @Override
   public String toString() {
     return "Params{"
```
satish-mittal, line 69: add booleanParams in `toString()` also

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -106,6 +106,9 @@ private PreparedStatement buildPreparedStatement(String statement, Params params
       params
           .getByteStringParams()
           .forEach((i, b) -> preparedStatement.setString(i, Hex.encodeHexString(b.toByteArray())));
+      params
```
satish-mittal, line 109: Might be good to add a comment mentioning why we store boolean type as String

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryRequestBuilderUtils.java
```diff
@@ -120,6 +120,10 @@ public static Filter createInFilter(String column, List<String> values) {
     return createFilter(column, Operator.IN, createStringArrayLiteralValueExpression(values));
   }
 
+  public static Filter createBooleanInFilter(String column, List<Boolean> values) {
```
satish-mittal, line 123:
```suggestion
  public static Filter createInFilter(String column, List<Boolean> values) {
```
This can be an overloaded method since the signature is different from existing one.

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryRequestBuilderUtils.java
```diff
@@ -120,6 +120,10 @@ public static Filter createInFilter(String column, List<String> values) {
...
+    return createFilter(column, Operator.IN, createBooleanArrayLiteralValueExpression(values));
```
satish-mittal, line 124: Basic question: what does it even mean to have a filter where rhs is a list of boolean values? Given that boolean can have only 2 valid values, expected value should be exactly one of them. 
- If both were to be passed together (e.g. `IN(true, false)`), that effectively means boolean value doesn't matter (and hence don't even pass the filter).
- If same value is passed multiple times (e.g. `IN(true, true, true)`), that is equivalent to just `EQ(true)`.

aaron-steinfeld: I think it's mostly just a consistency question - because yes, you're right it doesn't unlock any functionality that isn't there with either 0 or 1 other filters (unless it's a nullable boolean - which we do actually have). But either way by allowing it, we allow certain logic paths to not have to special case boolean values and translate different attribute type filters differently.

satish-mittal: Not sure about nullable boolean, but in both cases above it could help optimise the filters. In first case, we can just skip passing the filter. In second case, we can use EQ which is faster than IN.

aaron-steinfeld: Yeah - no issues if we want to translate them at this layer - I just wouldn't force the optimization on the caller.

skjindal93:
We tried to add a similar support before https://github.com/hypertrace/query-service/pull/170, and reverted back in https://github.com/hypertrace/query-service/pull/171

@saxenakshitiz had some concerns around supporting IN clause filter on boolean values, though it was for a different use case

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/Params.java
kotharironak: Done.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -512,11 +512,22 @@ private String convertLiteralToString(LiteralConstant literal, Params.Builder pa
         ret = QUESTION_MARK;
         paramsBuilder.addStringParam("null");
         break;
+      case BOOLEAN_ARRAY:
```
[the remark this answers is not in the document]
kotharironak: Done.

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryRequestBuilderUtils.java
kotharironak: In Java, when you use methods with generic types like List<Boolean> and List<String, the compiler can't distinguish them because they both appear as plain List without specific type information due to type erasure.

- https://www.tutorialspoint.com/java_generics/java_generics_no_overload.htm
- https://www.baeldung.com/java-type-erasure

kotharironak:
> We tried to add a similar support before #170, and reverted back in #171
> 
> @saxenakshitiz had some concerns around supporting IN clause filter on boolean values, though it was for a different use case

Why was it reverted? What was it breaking?

kotharironak:
> > We tried to add a similar support before #170, and reverted back in #171
> > @saxenakshitiz had some concerns around supporting IN clause filter on boolean values, though it was for a different use case
> 
> Why was it reverted back? was it breaking something?

query-service-impl/src/test/java/org/hypertrace/core/query/service/QueryRequestBuilderUtils.java
kotharironak: If a simple boolean type suffice, there may not be a need for an API interface with a BOOLEAN_ARRAY operator. But, I guess it was added for convenient for nullable boolean having three buckets (true, false, null).

In that case, do we expect the query like below:

```
childFilter {
  childFilter {
    lhs {
      attribute_expression {
        attributeId: "EVENT.isApiExternal"
        alias: "isApiExternal"
      }
    }
    operator: IN
    rhs {
      literal {
        value {
          valueType: BOOLEAN_ARRAY
          boolean_array: null
          boolean_array: true
        }
      }
    }
  }
}
```
If yes, we would need to ensure that we store all boolean fields as string fields in Pinot. We would need to enforce this during pinot schema creation for a view. And we have to go with `value as string` approach so PR #170 

If we store them as boolean type in pinot, nullable boolean will be consider as boolean value `false`. That will loose `null` bucket.

Regarding performance, there won't be much difference if data type is boolean in Pinot. For boolean data type, using IN (true, true, true) is same as EQ(true). And for IN(true, false), the suggestion will make difference if the column is exclusively used in filtering. We can make these change. But, looks like we will have to go with `string` base approach to include `null` bucket.

aaron-steinfeld: > If yes, we would need to ensure that we store all boolean fields as string fields in Pinot. 

What's our state today? If we are mixed, then we need to include null support including IN clauses. If we're consistently boolean, seems like we can make a few optimizations like dropping the in(true, false) case. I wouldn't build to support the hypothetical nullable boolean if we're not using it.

kotharironak: >What's our state today? 

Our current system has a mixed approach for pinot schema regarding Avro boolean data types. Except one view,  Avro booleans are stored as strings in pinot. This change in schema behavior is a result of upgrade of Pinot client library. Previously, the it treated booleans as strings, but now it treats them as booleans.

The pinot query works as string input for boolean field. Below are valid queries:
```
-- value as string
select api_is_external, api_id from bareSpanEventView
where start_time_millis > 1693470600000 and start_time_millis < 1693474200000
and is_external IN ('true')
limit 10
```
or
```
-- value as string
select api_is_external, api_id from bareSpanEventView
where start_time_millis > 1693470600000 and start_time_millis < 1693474200000
and is_external IN ('null')
limit 10
```

We are already treating boolean as string in [QS](https://github.com/hypertrace/query-service/blob/559f23bdbc90e5ab71af268a55d95b932231ffc7/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java#L499). So, applying same for boolean_array.

satish-mittal (approved):

skjindal93 (approved):

PR #195, opened 2023-09-19
chore: added error log

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotClientFactory.java
```diff
@@ -84,15 +87,24 @@ public ResultSetGroup executeQuery(String statement, Params params) {
...
-    public Future<ResultSetGroup> executeQueryAsync(String statement, Params params) {
-      /*
-      PreparedStatement preparedStatement = buildPreparedStatement(statement, params);
-      return preparedStatement.executeAsync();
-      */
-      return connection.executeAsync(new Request(SQL_FORMAT, resolveStatement(statement, params)));
+      try {
+        return connection.execute(new Request(SQL_FORMAT, resolveStatement(statement, params)));
+      } catch (Exception ex) {
+        if (rateLimiter.tryAcquire()) {
```
satish-mittal, line 93: do we really need to rate limit logs in case of query execution failure? should be quite less occurrence.

aaron-steinfeld: Agree - try to reduce reliance on rate limiter wherever possible to avoid contention (since we're acquiring a lock here).

In general, see comment on gateway PR as it applies here too: https://github.com/hypertrace/gateway-service/pull/179/files#r1329990699

saxenakshitiz: Error is already logged in one place, so added more details there

kotharironak (approved):

PR #214, opened 2023-11-10
chore: handle case of empty child filter

saxenakshitiz:


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -136,10 +136,16 @@ private String convertFilterToString(
...
       builder.append("( ");
       for (Filter childFilter : filter.getChildFilterList()) {
-        builder.append(delim);
-        builder.append(convertFilterToString(childFilter, paramsBuilder, executionContext));
-        builder.append(" ");
-        delim = operator + " ";
+        // child filter should either have further child filters or should have LHS and RHS
+        // only then child filter is converted to string
+        // this check avoids converting empty filters to string
+        if (!childFilter.getChildFilterList().isEmpty()
```
satish-mittal, line 142: These checks should go as part of query transformation that happens before actual query execution. See `AbstractQueryTransformation.transform()`. There is already transformFilter() that skips certain defaults.

kotharironak: +1

query-service-impl/src/test/java/org/hypertrace/core/query/service/AbstractServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.stream.Collectors;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+public abstract class AbstractServiceTest<TQueryServiceRequestType extends GeneratedMessageV3> {
+
+  private static final String QUERY_SERVICE_TEST_REQUESTS_DIR = "test-requests";
+
+  protected TQueryServiceRequestType readQueryServiceRequest(String fileName) {
```
kotharironak, line 17: What is `T`stands for? What about just having `QueryServiceRequestType` Is it how it was named in `proto`?

query-service-impl/src/test/java/org/hypertrace/core/query/service/AbstractServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+    String resourceFileName =
+        createResourceFileName(
+            QUERY_SERVICE_TEST_REQUESTS_DIR,
+            fileName); // QUERY_SERVICE_TEST_REQUESTS_DIR + FILE_SEPARATOR + getTestSuiteName() +
```
kotharironak, line 21: nit: shall we remove this comment? The function `createResourceFileName` internally explain this.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
saxenakshitiz: done

query-service-impl/src/test/java/org/hypertrace/core/query/service/AbstractServiceTest.java
saxenakshitiz: done

query-service-impl/src/test/java/org/hypertrace/core/query/service/AbstractServiceTest.java
saxenakshitiz: I think it is to denote a generic class type. It is same as definer here by you - https://github.com/hypertrace/gateway-service/blob/1997536bcdddfc4d00aed3af7392fc1bb2ecdf91/gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/AbstractServiceTest.java#L45

satish-mittal (approved):

kotharironak (approved):

PR #215, opened 2023-11-28
fix: handles default case for string array IN filter for pinot handlers

kotharironak:
## Description
The default value for string column is `null` in pinot - https://docs.pinot.apache.org/configuration-reference/schema
So, there are cases when query can only returns `null` group. These cases a follow up query with empty clause as shown in the below example:

```
Select dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','300:SECONDS'), span_name, COUNT(*) FROM spanEventView WHERE customer_id = '__default' 
AND start_time_millis >= 1701052800000 AND start_time_millis < 1701138900000 
AND ( user_roles IN ('null') ) ) ) 
GROUP BY dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','300:SECONDS'), user_roles 
limit 1000
```

It could possible that given time limit window, we have only one group `null`. 

This PR, handles the above default scenario.

### Testing
Added Unit test.

### Checklist:
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
@@ -450,7 +450,9 @@ class QueryRequestToPinotSQLConverter {
     String ret = null;
     switch (value.getValueType()) {
       case STRING_ARRAY:
-        ret = buildArrayValue(value.getStringArrayList(), paramsBuilder::addStringParam);
+        List<String> values =
+            value.getStringArrayList().size() > 0 ? value.getStringArrayList() : List.of("null");
+        ret = buildArrayValue(values, paramsBuilder::addStringParam);
         break;
       case BYTES_ARRAY:
         ret = buildArrayValue(value.getBytesArrayList(), paramsBuilder::addByteStringParam);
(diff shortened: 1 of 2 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/QueryRequestToPinotSQLConverter.java
```diff
@@ -450,7 +450,9 @@ private String convertLiteralToString(LiteralConstant literal, Params.Builder pa
     String ret = null;
     switch (value.getValueType()) {
       case STRING_ARRAY:
-        ret = buildArrayValue(value.getStringArrayList(), paramsBuilder::addStringParam);
+        List<String> values =
+            value.getStringArrayList().size() > 0 ? value.getStringArrayList() : List.of("null");
```
varsha-abhinandan, line 454: Why is gateway service converting null string to empty ? Shouldn't be fix that instead of expecting every converter to handle this ?

saxenakshitiz: 'null' is special case on how it is stored in Pinot and needs to be converted to an appropriate value. Isn't it?

saxenakshitiz, line 454: why are we handling special case only for string array? There are other array types as well. We don't need to handle this case for them?

kotharironak:
Closing this with respect : https://github.com/hypertrace/gateway-service/pull/186

PR #220, opened 2024-02-01
chore: fix postgres query for array type column

saxenakshitiz:


aaron-steinfeld (dismissed):

query-service/src/integrationTest/java/org/hypertrace/core/query/service/htqueries/HTPinotQueriesTest.java
```diff
@@ -95,7 +95,7 @@ public static void setup() throws Exception {
     kafkaZk.start();
 
     pinotServiceManager =
-        new GenericContainer<>(DockerImageName.parse("hypertrace/pinot-servicemanager:main"))
+        new GenericContainer<>(DockerImageName.parse("hypertrace/pinot-servicemanager:0.6.12"))
```
saxenakshitiz, line 98: Filed a tracking ticket to make it work it with main version. Once fixed, will revert this change

.trivyignore
```diff
@@ -1 +1,2 @@
-
+CVE-2023-6246 exp:2024-02-29
```
aaron-steinfeld, line 1: These shouldn't be needed - I believe we handled them globally

PR #224, opened 2024-07-22
ENG-46252 : Pinot time range filtering support

anujgoyal1:


kotharironak:
Can we add some tests?

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -82,6 +82,18 @@ service.config = {
       requestHandlerInfo = {
         tenantColumnName = tenant_id
         startTimeAttributeName = "EVENT.startTime"
+#         additionalTenantFilters = [
+#           {
+#             "tenantId" : "testTenant",
+#             "timeRangeAndFilters" : [
+#               {
+#                 "start_time_millis" : 1,
+#                 "endTimeMillis" : 1,
```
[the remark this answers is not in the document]
anujgoyal1: Yes, but I'm thinking of making `start_time_millis` as starTimeMillis to keep it consistent with the other keys

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -82,6 +82,18 @@ service.config = {
       requestHandlerInfo = {
         tenantColumnName = tenant_id
         startTimeAttributeName = "EVENT.startTime"
+#         additionalTenantFilters = [
```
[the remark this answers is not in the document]
anujgoyal1: since it's already under a table scope, I think we not need add another scope here. 
I don't think we need to further extend it to query scope(Its not even available in query as we discussed)

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -82,6 +82,18 @@ service.config = {
...
+#           {
+#             "tenantId" : "testTenant",
+#             "timeRangeAndFilters" : [
```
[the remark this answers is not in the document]
anujgoyal1: yes

query-service-impl/src/main/java/org/hypertrace/core/query/service/ExecutionContext.java
```diff
@@ -3,9 +3,6 @@
 import static org.hypertrace.core.query.service.QueryFunctionConstants.QUERY_FUNCTION_COUNT;
 import static org.hypertrace.core.query.service.QueryRequestUtil.getAlias;
 import static org.hypertrace.core.query.service.QueryRequestUtil.getLogicalColumnName;
-import static org.hypertrace.core.query.service.api.Expression.ValueCase.ATTRIBUTE_EXPRESSION;
```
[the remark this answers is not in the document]
anujgoyal1: reverted

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceImpl.java
```diff
@@ -97,8 +103,23 @@ private Observable<ResultSetChunk> executeTransformedRequest(
...
             handler -> {
               handler.getTimeFilterColumn().ifPresent(context::setTimeFilterColumn);
-              return handler.handleRequest(transformedRequest, context);
+              return handler.handleRequest(
+                  handler.applyAdditionalFilters(transformedRequest, context), context);
             })
         .lift(chunkRows(context.getResultSetMetadata()));
   }
+
+  public static void main(String[] args) {
```
kotharironak, line 112: Can you move this to test or remove it?

query-service-impl/src/test/resources/application.conf
```diff
@@ -39,6 +39,18 @@ service.config = {
...
         slowQueryThresholdMs = 0
         percentileAggFunction = "PERCENTILE"
+        additionalTenantFilters = [
+          {
+            "tenantId" = "__default"
+            "timeRangeAndFilters" = [
+              {
+                "startTimeMillis" = 1
+                "endTimeMillis" = 2
+                "filter" : "{\r\n  \"lhs\": {\r\n    \"columnIdentifier\": {\r\n      \"columnName\": \"SERVICE.name\"\r\n    }\r\n  },\r\n  \"operator\": \"EQ\",\r\n  \"rhs\": {\r\n    \"literal\": {\r\n      \"value\": {\r\n        \"string\": \"dummyService\"\r\n      }\r\n    }\r\n  }\r\n}"
```
kotharironak, line 49: Q: Do we need `\r\n`? If we just put a valid single line json string, would it work?

query-service-impl/src/test/resources/application.conf
```diff
@@ -389,6 +412,16 @@ service.config = {
       requestHandlerInfo = {
         tenantColumnName: customer_id
         startTimeAttributeName = "EVENT.startTime"
+        additionalTenantFilters = [
+                  {
+                    "tenantId" = "__default"
+                    "timeRangeAndFilters" = [
+                      {
+                        "filter" : "{\r\n  \"lhs\": {\r\n    \"columnIdentifier\": {\r\n      \"columnName\": \"SERVICE.name\"\r\n    }\r\n  },\r\n  \"operator\": \"EQ\",\r\n  \"rhs\": {\r\n    \"literal\": {\r\n      \"value\": {\r\n        \"string\": \"dummyService\"\r\n      }\r\n    }\r\n  }\r\n}"
```
kotharironak, line 420: Do we need `\r\n`? If we just put a single string, would it work?

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+        endTimeMillis = Optional.empty();
+      }
+      if (config.hasPath(FILTER_CONFIG_PATH)) {
+        this.filter = deserializeFilter(config.getString(FILTER_CONFIG_PATH));
+      } else {
+        filter = Optional.empty();
+      }
+    }
+
+    private Optional<Filter> buildTimeRangeAndFilters(Optional<String> timeRangeAttributeOptional) {
```
kotharironak, line 88: nit: timeRangeAttributeOptional -> timeRangeAttribute

query-service-impl/src/test/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandlerTest.java
```diff
@@ -141,6 +143,119 @@ public void testCanHandle() {
...
+      if (!isPinotConfig(config)) {
+        continue;
+      }
+
+      PinotBasedRequestHandler handler =
+          new PinotBasedRequestHandler(
+              config.getString("name"), config.getConfig("requestHandlerInfo"));
+
+      // domain-service-handler only has filter and no time range
+      if (config.getString("name").equals("domain-service-handler")) {
```
kotharironak, line 158: Is this test using the correct handler - `domain-service-handler`? For this, the original request should be the same after applying filters, right?

kotharironak: I think, it should have been - `trace-view-handler`, right?

anujgoyal1: domain-event-view only has filter.
span-event-view only has time range 
trace-view has both time range and filter.

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+      Filter.Builder filterBuilder = Filter.newBuilder();
+      filterBuilder.setOperator(Operator.OR);
+      if (timeRangeAttributeOptional.isPresent()) {
+        if (this.startTimeMillis.isPresent() && this.endTimeMillis.isPresent()) {
+          filterBuilder.addChildFilter(
+              QueryRequestUtil.createLongFilter(
+                  timeRangeAttributeOptional.get(), Operator.GT, this.endTimeMillis.get()));
```
kotharironak, line 95: Q: Is this check correct? Should `endTime` be compared with the greater than operator (I don't remember the text you had in your sublime, so just re-checking)?

query-service-impl/src/test/resources/application.conf
anujgoyal1: Yeah, it would work. Was just added by the json string escaper I was using, got rid of it.

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
anujgoyal1: Yes this is how it's supposed to be as it's an or filter 
This is what we discussed.

```
t3....   t1  t2 ....t4


((s >= t2 OR s <= t1) OR (API NOT IN (api1, api2)) AND (t3, t4))
```

query-service-impl/src/test/resources/application.conf
anujgoyal1: removed

query-service-impl/src/main/java/org/hypertrace/core/query/service/QueryServiceImpl.java
anujgoyal1: aah, my bad. removed

kotharironak (approved):

kotharironak (dismissed):

kotharironak (dismissed):

query-service-impl/src/main/java/org/hypertrace/core/query/service/RequestHandler.java
```diff
@@ -27,4 +27,9 @@ public interface RequestHandler {
 
   /** Handle the request and add rows to the collector. */
   Observable<Row> handleRequest(QueryRequest request, ExecutionContext executionContext);
+
+  default QueryRequest applyAdditionalFilters(
```
satish-mittal, line 31: There is already an existing step of query transformation. Look at `QueryTransformation`. This step of adding filters should go there.

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+                      timeRangeAndFilter.buildTimeRangeAndFilters(startTimeAttributeName))
+              .filter(Optional::isPresent)
+              .map(Optional::get)
+              .collect(Collectors.toList());
+    }
+  }
+
+  @lombok.Value
+  @NonFinal
+  private static class TimeRangeAndFilter {
```
satish-mittal, line 64:
```suggestion
  private static class FiltersForTimeRange {
```
The existing name `TimeRangeAndFilter ` is quite confusing. It sounds like it is carrying 2 unrelated concepts together: timeRange and filters.

query-service/src/main/resources/configs/common/application.conf
satish-mittal, line 85:
```suggestion
#         tenantScopedFilters = [
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+      if (config.hasPath(FILTER_CONFIG_PATH)) {
+        this.filter = deserializeFilter(config.getString(FILTER_CONFIG_PATH));
+      } else {
+        filter = Optional.empty();
+      }
+    }
+
+    private Optional<Filter> buildTimeRangeAndFilters(Optional<String> timeRangeAttribute) {
+      Filter.Builder filterBuilder = Filter.newBuilder();
+      filterBuilder.setOperator(Operator.OR);
```
satish-mittal, line 90: why OR? don't we want the filter to be: time > startTime and time < endTime?

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+              .collect(
+                  Collectors.toMap(
+                      tenantFilters -> tenantFilters.tenantId,
+                      tenantFilters -> tenantFilters.filters));
+    } else {
+      this.tenantToAdditionalFiltersMap = Collections.emptyMap();
+    }
+  }
+
+  public List<Filter> getAdditionalFiltersForTenant(String tenantId) {
```
satish-mittal, line 36: We should definitely add test cases for this new class `AdditionalHandlerFiltersConfig`.

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import lombok.experimental.NonFinal;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.query.service.api.Filter;
+import org.hypertrace.core.query.service.api.Operator;
+
+@Slf4j
+public class AdditionalHandlerFiltersConfig {
```
satish-mittal, line 18:
```suggestion
public class HandlerScopedFiltersConfig {
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
```diff
@@ -0,0 +1,128 @@
...
+  private static final String ADDITIONAL_TENANT_FILTERS_CONFIG_KEY = "additionalTenantFilters";
```
satish-mittal, line 19:
```suggestion
  private static final String ADDITIONAL_TENANT_FILTERS_CONFIG_KEY = "tenantScopedFilters";
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
anujgoyal1: https://github.com/hypertrace/query-service/pull/224#discussion_r1689556713

query-service-impl/src/test/resources/application.conf
```diff
@@ -39,6 +39,18 @@ service.config = {
         startTimeAttributeName = "Trace.start_time_millis"
         slowQueryThresholdMs = 0
         percentileAggFunction = "PERCENTILE"
+        additionalTenantFilters = [
```
satish-mittal, line 42: There are multiple problems in the way this new config has been defined:
- It is  is very confusing. Only after going through the code one can understand.
- There is a hardcoded assumption that `startTimeAttributeName` will act as the lhs key for the 2 values below. Makes it very specific and no room for generalising.
- It is hardcoded with `OR` operator.

I would prefer it being defined as follows:
```
additionalHandlerConfigs = [
  operator = "OR"
  {
    key = "Trace.start_time_millis"
    operator = LT
    value = "max_cutoff"
  },
  {
    key = "Trace.start_time_millis"
    operator = GE
    value = "min_cutoff"
  },
  filter : {
    columnIdentifiier: ...
  }
]
```

kotharironak (commented):
lgtm

kotharironak (approved):

query-service-impl/src/test/resources/application.conf
kotharironak: Would `operator = "AND"` possible here?

kotharironak: We would like to apply the time range to `startTimeAttributeName`. Any other attribute can go to the normal filter expression.

The given time range indicates that we would like to apply a filter within that specified range. is that not clear?

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
anujgoyal1: All scenarios for this config is already covered in PinotHanderTest since we're taking input from application.conf there.

query-service-impl/src/main/java/org/hypertrace/core/query/service/RequestHandler.java
anujgoyal1: We started from this, then realised we don't get scope information in query request.
If we apply a filter on Transformation step, it'll be applied on all the scopes and would work as global filter. Which is wrong behaviour
Query Service decides the scope and it's handler  based of the query. Hence pushed down these filters to handler

query-service-impl/src/main/java/org/hypertrace/core/query/service/AdditionalHandlerFiltersConfig.java
anujgoyal1: Added config tests too

kotharironak (approved):
LGTM

PR #227, opened 2024-09-27
ENG-50887: Mask value using a masking config

siddhant2001:
- [x] Write tests and verify correctness

### DIFF
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
@@ -4,9 +4,11 @@ import com.typesafe.config.Config;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.Value;
 import lombok.experimental.NonFinal;
@@ -30,9 +32,9 @@ public class HandlerScopedMaskingConfig {
     }
   }
 
-  public List<String> getMaskedAttributes(ExecutionContext executionContext) {
+  public Set<String> getMaskedAttributes(ExecutionContext executionContext) {
     String tenantId = executionContext.getTenantId();
-    List<String> maskedAttributes = new ArrayList<>();
+    HashSet<String> maskedAttributes = new HashSet<>();
     if (!tenantToTimeRangeMaskedAttributes.containsKey(tenantId)) {
       return maskedAttributes;
     }
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
@@ -14,7 +14,7 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class HandlerScopedMaskingConfig {
-  private static final String TENANT_SCOPED_MASKS_CONFIG_KEY = "tenantScopedMaskingCriteria";
+  private static final String TENANT_SCOPED_MASKS_CONFIG_KEY = "tenantScopedMaskingConfig";
   private Map<String, List<TimeRangeToMaskedAttributes>> tenantToTimeRangeMaskedAttributes =
       Collections.emptyMap();
 
@@ -62,7 +62,8 @@ public class HandlerScopedMaskingConfig {
   @NonFinal
   static class TenantMaskingConfig {
     private static final String TENANT_ID_CONFIG_KEY = "tenantId";
-    private static final String TIME_RANGE_AND_MASK_VALUES_CONFIG_KEY = "timeRangeAndMaskValues";
+    private static final String TIME_RANGE_AND_MASK_VALUES_CONFIG_KEY =
+        "timeRangeToMaskedAttributes";
     String tenantId;
     List<TimeRangeToMaskedAttributes> timeRangeToMaskedAttributes;
 
--- a/query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
+++ b/query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
@@ -33,11 +33,12 @@ public class HandlerScopedMaskingConfig {
     }
   }
 
-  public void parseColumns(ExecutionContext executionContext) {
+  public HashMap<String, String> getColumnToMaskedValuesMap(ExecutionContext executionContext) {
     String tenantId = executionContext.getTenantId();
-    maskedValue.clear();
+    HashMap<String, String> columnMaskedValue = new HashMap<>();
+    //    maskedValue.clear();
     if (!tenantToMaskValuesMap.containsKey(tenantId)) {
-      return;
+      return columnMaskedValue;
     }
 
     Optional<QueryTimeRange> queryTimeRange = executionContext.getQueryTimeRange();
@@ -57,10 +58,12 @@ public class HandlerScopedMaskingConfig {
         Map<String, String> attributeToMaskedValue =
             timeRangeAndMasks.maskValues.attributeToMaskedValue;
         for (String attribute : attributeToMaskedValue.keySet()) {
-          maskedValue.put(attribute, attributeToMaskedValue.get(attribute));
+          columnMaskedValue.put(attribute, attributeToMaskedValue.get(attribute));
         }
       }
     }
+
+    return columnMaskedValue;
   }
 
   private static boolean isTimeRangeOverlap(
@@ -81,14 +84,6 @@ public class HandlerScopedMaskingConfig {
       }
     }
     return timeRangeOverlap;
-  }
-
-  public boolean shouldMask(String attributeName) {
-    return this.maskedValue.containsKey(attributeName);
-  }
-
-  public String getMaskedValue(String attributeName) {
-    return this.maskedValue.get(attributeName);
   }
 
   @Value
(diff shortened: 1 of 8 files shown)


query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
```diff
@@ -0,0 +1,158 @@
...
+          shouldMaskAttribute.put(attribute, true);
+          maskedValue.put(attribute, attributeToMaskedValue.get(attribute));
+        }
+      }
+    }
+  }
+
+  private static boolean isTimeRangeOverlap(
+      MaskValuesForTimeRange timeRangeAndMasks, Instant queryStartTime, Instant queryEndTime) {
+    boolean timeRangeOverlap = true;
```
kotharironak, line 71: Default should be false, right?

query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
```diff
@@ -0,0 +1,158 @@
...
+
+    if (timeRangeAndMasks.getStartTimeMillis().isPresent()) {
+      Instant startTimeInstant = Instant.ofEpochMilli(timeRangeAndMasks.getStartTimeMillis().get());
+      if (startTimeInstant.isBefore(queryStartTime) || startTimeInstant.isAfter(queryEndTime)) {
+        timeRangeOverlap = false;
+      }
+
+      Instant endTimeInstant = Instant.ofEpochMilli(timeRangeAndMasks.getStartTimeMillis().get());
```
kotharironak, line 79: should be `getEndTimeMillis`?

kotharironak: I guess, is this what we are looking as function?
```
if (timeRangeAndMasks.getStartTimeMillis().isPresent()) {
      
      Instant startTimeInstant = Instant.ofEpochMilli(timeRangeAndMasks.getStartTimeMillis().get());
      Instant endTimeInstant = Instant.ofEpochMilli(timeRangeAndMasks.getEndTimeMillis().get());

      if (!(startTimeInstant.isAfter(queryEndTime) || endTimeInstant.isBefore(queryStartTime))) {
        return true;
      }
    }
    return false;
```

query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
```diff
@@ -0,0 +1,158 @@
...
+                  Collectors.toMap(
+                      tenantFilters -> tenantFilters.tenantId,
+                      tenantFilters -> tenantFilters.maskValues));
+    } else {
+      this.tenantToMaskValuesMap = Collections.emptyMap();
+    }
+  }
+
+  public void parseColumns(ExecutionContext executionContext) {
+    shouldMaskAttribute.clear();
```
kotharironak, line 38: It seems like the state is maintained per request, but we should only test against the timeRange condition.

Pre-compute using config:
- tenantId -> List of timeRanges
- timeRange -> set of attributes
- timeRange -> map(attributeId, maskValue)

During response processing:
- Check if any time range matches.
- Pick the first match (or should we apply UNION?).
   - If UNION is used, and the same attribute is present in two time ranges with different mask values, which one should we consider? I guess any Should be fine.

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -94,6 +94,27 @@ service.config = {
 #             ]
 #            }
 #         ]
+        tenantScopedMaskingCriteria = [
+          {
+            "tenantId": "testTenant",
+            "timeRangeAndMaskValues": [
```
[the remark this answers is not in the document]
siddhant2001: Yes, each timeRange condition will have it's own masking values

query-service/src/main/resources/configs/common/application.conf
```diff
@@ -94,6 +94,27 @@ service.config = {
 #             ]
 #            }
 #         ]
+        tenantScopedMaskingCriteria = [
```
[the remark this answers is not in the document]
siddhant2001: Yes done, you've reviewed an older commit

query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.stream.Collectors;
+import lombok.Value;
+import lombok.experimental.NonFinal;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class HandlerScopedMaskingConfig {
+  private static final String TENANT_SCOPED_MASKS_CONFIG_KEY = "tenantScopedMaskingCriteria";
+  private final Map<String, List<MaskValuesForTimeRange>> tenantToMaskValuesMap;
+  private HashMap<String, Boolean> shouldMaskAttribute = new HashMap<>();
```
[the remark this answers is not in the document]
siddhant2001: AttributeId is the key to both.
Removing shouldMaskAttribute as it is not needed.

query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
siddhant2001: The following conditionals check for no overlap, i.e. they set the timeRangeOverlap to false. This statement is correct.

query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
siddhant2001: Right, fixed the condition.

query-service-impl/src/main/java/org/hypertrace/core/query/service/HandlerScopedMaskingConfig.java
siddhant2001: This is how I've done it. 
In case of attribute in multiple time ranges, I choose any value.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotResultAnalyzer.java
```diff
@@ -149,4 +157,12 @@ String getDataFromRow(int rowIndex, String logicalName) {
     }
     return result;
   }
+
+  String getLogicalNameFromColIdx(Integer colIdx) {
+    if (indexToLogicalName.containsKey(colIdx)) {
+      return indexToLogicalName.get(colIdx);
+    }
+
```
kotharironak, line 165: nit: remove additional space.

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -58,6 +59,10 @@ public class PinotBasedRequestHandler implements RequestHandler {
   private static final String START_TIME_ATTRIBUTE_NAME_CONFIG_KEY = "startTimeAttributeName";
   private static final String SLOW_QUERY_THRESHOLD_MS_CONFIG = "slowQueryThresholdMs";
 
+  private static final String MASKED_VALUE = "*";
+  // This is how empty list is represented in Pinot
+  private static final String PINOT_EMPTY_LIST = "[\"\"]";
```
kotharironak, line 64: `PINOT_EMPTY_LIST` -> `ARRAY_TYPE_MASKED_VALUE`
`MASKED_VALUE`-> `DEFAULT_MASKED_VALUE`

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -493,21 +500,26 @@ private Filter rewriteLeafFilter(
     return queryFilter;
   }
 
-  Observable<Row> convert(ResultSetGroup resultSetGroup, LinkedHashSet<String> selectedAttributes) {
+  Observable<Row> convert(ResultSetGroup resultSetGroup, ExecutionContext executionContext) {
+    LinkedHashSet<String> selectedAttributes = executionContext.getSelectedColumns();
```
kotharironak, line 504: can you move this also inside `resultSetGroup.getResultSetCount() > 0`?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
```diff
@@ -588,22 +605,36 @@ private void handleAggregationAndGroupBy(
...
+          PinotResultAnalyzer.create(resultSet, selectedAttributes, viewDefinition);
       for (int rowIdx = 0; rowIdx < resultSet.getRowCount(); rowIdx++) {
         Builder builder;
         builder = Row.newBuilder();
         rowBuilderList.add(builder);
 
-        for (int colIdx = 0; colIdx < resultSet.getColumnCount(); colIdx++) {
+        for (int colIdx = 0, logicalColIdx = 0;
+            colIdx < resultSet.getColumnCount();
+            colIdx++, logicalColIdx++) {
```
kotharironak, line 624: Why do we need `logicalColIdx`?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotResultAnalyzer.java
```diff
@@ -149,4 +158,8 @@ String getDataFromRow(int rowIndex, String logicalName) {
     }
     return result;
   }
+
+  Optional<String> getLogicalNameFromColIdx(Integer colIdx) {
+    return Optional.ofNullable(indexToLogicalName.get(colIdx));
```
kotharironak, line 163: On what scenario, can it be null?

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java
siddhant2001: There can be multiple map fields. When creating the idxToLogical name map, a map field would only increment the counter by one. Here each map field increments colIdx by 2. That's why I have created a new variable, which is incremented only once even if we go through a maps 2 columns (key and value)

query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotResultAnalyzer.java
siddhant2001: You're right, it shouldn't be

kotharironak (dismissed):
LGTM

kotharironak (approved):

